PR #6, opened 2020-06-22
Ensure fixed shape array when history!=0, fix tests

lucabergamini:
Now `BoxRasterizer` always returns the same number of channels. If the history is not there (e.g. beginning of a scene) black channels are concatenated to pad the raster output.

It was crashing pytorch because of mismatched sizes in the batch

### DIFF
--- a/l5kit/l5kit/rasterization/box_rasterizer.py
+++ b/l5kit/l5kit/rasterization/box_rasterizer.py
@@ -119,8 +119,8 @@ class BoxRasterizer(Rasterizer):
         )
 
         # this ensures we always end up with fixed size arrays, +1 is because current time is also in the history
-        agents_im = np.zeros((*self.raster_size, self.history_num_frames + 1), dtype=np.uint8)
-        ego_im = np.zeros((*self.raster_size, self.history_num_frames + 1), dtype=np.uint8)
+        agents_images = np.zeros((*self.raster_size, self.history_num_frames + 1), dtype=np.uint8)
+        ego_images = np.zeros((*self.raster_size, self.history_num_frames + 1), dtype=np.uint8)
 
         for i, frame in enumerate(history_frames):
             agents = filter_agents_by_frame(all_agents, frame)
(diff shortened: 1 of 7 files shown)


l5kit/l5kit/rasterization/box_rasterizer.py
```diff
@@ -115,8 +118,9 @@ def rasterize(
             ego_center_in_image_ratio=self.ego_center,
         )
 
-        agents_im = []
-        ego_im = []
+        # this ensure we always end up with fixed size arrays, +1 is because cur time in also in history
```
laky, line 121:
```suggestion
        # this ensures we always end up with fixed size arrays, +1 is because current time is also in the history
```

Nit: typos

l5kit/l5kit/rasterization/box_rasterizer.py
```diff
@@ -137,11 +141,11 @@ def rasterize(
                     im = draw_boxes(self.raster_size, world_to_image_space, np.append(agents, av_agent), 255)
```
laky, line 141: Can we make this name a bit more descriptive, please? Something like `rasterized_agent_boxes_frame`

`im_ego` => rasterized_ego_box_frame

Or something, it's a bit confusing to read now and understand what is im, agents_im, ego_im.

l5kit/l5kit/rasterization/box_rasterizer.py
```diff
@@ -137,11 +141,11 @@ def rasterize(
...
                     im_ego = draw_boxes(self.raster_size, world_to_image_space, agent_ego, 255)
 
-            agents_im.append(im)
-            ego_im.append(im_ego)
+            agents_im[..., i] = im
+            ego_im[..., i] = im_ego
```
laky, line 145: Can we do assignment to these in the loop and get rid of those temp variables? I think it would be cleaner to understand.

lucabergamini: I'll give the variable better names, but replacing with direct assignement is too error prone imo

PR #23, opened 2020-06-24
Lberg.add filter zarr

lucabergamini:
This PR add two script:
- `filter_zarr` filters a list of zarr based on a list of blacklists (indices of scene we want to delete)
- `get_zarr_coords.py` consume a list of zarrs and produce 2D coords of the AV trajectory and scene information


examples/visualisation/get_zarr_coords.py
```diff
@@ -0,0 +1,46 @@
+import numpy as np
```
oliver-scheel, line 1: is
```
import os
os.environ["L5KIT_DATA_FOLDER"] = "/Users/oliverscheel/L5KIT/data/"
```
needed or should be set anywas?

lucabergamini: I usually set it in the shell..It's only required to print the dataset tbh. Maybe I'll move the print into the zarr :D

PR #29, opened 2020-06-25
remove combine_scene, add print total steps

lucabergamini:
now scenes are consecutive, so joining them was impacting multi-threading performances


lucabergamini:
> Nice, so we start train & eval on chunked dataset?

The difference is that because scene are sorted, that function was melting together small scenes in few bigger ones. That could potentially harm multi-processing (and it also gives weird output in console because the number of scene doesn't match what you would expect)

PR #32, opened 2020-06-25
Lberg.speed up agents

lucabergamini:
Try to make agents faster. A report of the slowness causes will follow

### DIFF
--- a/l5kit/l5kit/data/filter.py
+++ b/l5kit/l5kit/data/filter.py
@@ -1,4 +1,4 @@
-from typing import Optional
+from typing import List, Optional
 
 import numpy as np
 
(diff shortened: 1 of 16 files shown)


l5kit/l5kit/sampling/agent_sampling.py
```diff
@@ -79,12 +80,16 @@ def generate_agent_sample(
         target for that state. If you sample near the end of a scene, this may contain zeroes.
 
     """
+    agent_index_start = frames[0]["agent_index_interval"][0]
+
     #  the history slice is ordered starting from the latest frame and goes backward in time., ex. slice(100, 91, -2)
     history_slice = get_history_slice(state_index, history_num_frames, history_step_size, include_current_state=True)
     history_frames = frames[history_slice]
+    history_agents = [agents[slice(*(hf["agent_index_interval"] - agent_index_start))] for hf in history_frames]
```
laky, line 88: Can we extract this into some function? It is a lot of copy paste. The worst part is that it is copy pasted into the test as well, so if you made a mistake in here, it would not get caught by the test. 

Ideally, write a test for this function as well to make sure you get all the frames as you expect.

l5kit/l5kit/dataset/ego.py
```diff
@@ -76,8 +76,9 @@ def get_frame(self, scene_index: int, state_index: int, track_id: Optional[int]
         """
         frame_interval = self.dataset.scenes[scene_index]["frame_index_interval"]
         frames = self.dataset.frames[frame_interval[0] : frame_interval[1]]
+        agents = self.dataset.agents[frames[0]["agent_index_interval"][0] : frames[-1]["agent_index_interval"][1]]
```
laky, line 79: This is quite tough to read. Could you do something like:

```suggestion
        start_agent_index = frames[0]["agent_index_interval"][0] 
        end_agent_index = frames[-1]["agent_index_interval"][1]
        agents = self.dataset.agents[start_agent_index:end_agent_index]
```

l5kit/l5kit/sampling/agent_sampling.py
lucabergamini: on it :)

l5kit/l5kit/data/filter.py
```diff
@@ -83,22 +83,35 @@ def filter_agents_by_track_id(agents: np.ndarray, track_id: int) -> np.ndarray:
...
-    agents = filter_agents_by_frame(all_agents, frame)
     try:
-        agent = filter_agents_by_track_id(agents, track_id)[0]
+        agent = filter_agents_by_track_id(agents_frame, track_id)[0]
         return agent
     except IndexError:  # no agent for track_id in this frame
         return None
+
+
+def get_frames_agents(frames: np.ndarray, agents: np.ndarray) -> List[np.ndarray]:
```
yye-lyft, line 105: @lucabergamini So what this function does is: given all the agents and frames, return only the agents from given frames?

yye-lyft: @lucabergamini I would make the docs more clear and have a better function name, otherwise LGTM

lucabergamini: > @lucabergamini I would make the docs more clear and have a better function name, otherwise LGTM
> `filter_agents_by_frames`? be consistent with other funcion name `filter_agents_by_frame`

done. Note that now the `filter_agents_by_frame` is not used in the code...we may want to remove it in the future @yye-lyft

PR #42, opened 2020-06-29
move print to zarr dataset

lucabergamini:
Move the dataset `__repr__` into the `ChunkedStateDataset`. Also put some sensible default for that class attributes

### DIFF
--- a/l5kit/l5kit/data/zarr_dataset.py
+++ b/l5kit/l5kit/data/zarr_dataset.py
@@ -123,7 +123,7 @@ opened.
             "Avg Frames per Scene",
             "Avg Agents per Frame",
             "Avg Scene Time (sec)",
-            "Avg Frame frequency"
+            "Avg Frame frequency",
         ]
         if len(self.agents) > 1:
             times = self.frames[1:10]["timestamp"] - self.frames[0:9]["timestamp"]
@@ -139,7 +139,7 @@ opened.
             len(self.frames) / max(len(self.scenes), 1e-9),
             len(self.agents) / max(len(self.frames), 1e-9),
             len(self.frames) / max(len(self.scenes), 1e-9) / 10,
-            frequency
+            frequency,
         ]
         table = PrettyTable(field_names=fields)
         table.float_format = ".2"
--- a/l5kit/l5kit/data/zarr_dataset.py
+++ b/l5kit/l5kit/data/zarr_dataset.py
@@ -1,5 +1,6 @@
 import numpy as np
 import zarr
+from prettytable import PrettyTable
 
 from .labels import LABELS
 
@@ -61,6 +62,9 @@ class ChunkedStateDataset:
         """
         self.key = key
         self.path = path
+        self.frames = np.empty(0, dtype=FRAME_DTYPE)
+        self.scenes = np.empty(0, dtype=SCENE_DTYPE)
+        self.agents = np.empty(0, AGENT_DTYPE)
 
         # Note: we still support only zarr. However, some functions build a new dataset so we cannot raise error.
         if ".zarr" not in self.path:
@@ -109,3 +113,27 @@ opened.
         self.frames = self.root[FRAME_ARRAY_KEY]
         self.agents = self.root[AGENT_ARRAY_KEY]
         self.scenes = self.root[SCENE_ARRAY_KEY]
+
+    def __repr__(self) -> str:
+        fields = [
+            "Num Scenes",
+            "Num Frames",
+            "Num Agents",
+            "Total Time (hr)",
+            "Avg Frames per Scene",
+            "Avg Agents per Frame",
+            "Avg Scene Time (sec)",
+        ]
+        values = [
+            len(self.scenes),
+            len(self.frames),
+            len(self.agents),
+            len(self.frames) / (10 * 3600),
+            len(self.frames) / len(self.scenes),
+            len(self.agents) / len(self.frames),
+            len(self.frames) / len(self.scenes) / 10,
+        ]
+        table = PrettyTable(field_names=fields)
+        table.float_format = ".2"
+        table.add_row(values)
+        return str(table)
--- a/l5kit/l5kit/data/zarr_dataset.py
+++ b/l5kit/l5kit/data/zarr_dataset.py
@@ -135,7 +135,7 @@ opened.
             len(self.scenes),
             len(self.frames),
             len(self.agents),
-            len(self.frames) / max(frequency * 3600, 1e-9),
+            len(self.frames) / max(frequency, 1e-9) / 3600,
             len(self.frames) / max(len(self.scenes), 1e-9),
             len(self.agents) / max(len(self.frames), 1e-9),
             len(self.frames) / max(len(self.scenes), 1e-9) / 10,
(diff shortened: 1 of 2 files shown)


l5kit/l5kit/data/zarr_dataset.py
```diff
@@ -109,3 +113,27 @@ def open(self, mode: str = "r", cached: bool = True, cache_size_bytes: int = int
...
+            "Total Time (hr)",
+            "Avg Frames per Scene",
+            "Avg Agents per Frame",
+            "Avg Scene Time (sec)",
+        ]
+        values = [
+            len(self.scenes),
+            len(self.frames),
+            len(self.agents),
+            len(self.frames) / (10 * 3600),
```
oliver-scheel, line 131: How about a param for the sampling frequency?

l5kit/l5kit/data/zarr_dataset.py
```diff
@@ -61,6 +62,9 @@ def __init__(self, path: str, key: str = ""):
         """
         self.key = key
         self.path = path
+        self.frames = np.empty(0, dtype=FRAME_DTYPE)
+        self.scenes = np.empty(0, dtype=SCENE_DTYPE)
+        self.agents = np.empty(0, AGENT_DTYPE)
```
yye-lyft, line 67: Minor: `self.agents = np.empty(0, dtype=AGENT_DTYPE)`

yye-lyft: I am not a fan of having three function to initialize dataset XD.
- `__init__`
- `initialize`
- `open`

l5kit/l5kit/data/zarr_dataset.py
lucabergamini: that information is not in the `.zarr` actually :(  It would be possible to compute it by reading the frames `timestamp`, but getting an average may slow the print function down a lot, because we need to read all the frames. We can also read the first N frames and compute their average, but I fear it won't be very indicative

yye-lyft: @lucabergamini Then a hard coded constant param for it, Like `FRAME_RATE=10  # Hz`

oliver-scheel: Ideally it would be great if we could read that out, but yes @yye-lyft , I meant that: In case we later change to a sampling frequency, we only have to modify it in one position.

yye-lyft: @lucabergamini

lucabergamini: Let me try to read it, if that doesn't succeed in an hour I would keep it for another time

l5kit/l5kit/data/zarr_dataset.py
```diff
@@ -109,3 +113,35 @@ def open(self, mode: str = "r", cached: bool = True, cache_size_bytes: int = int
...
+            "Num Frames",
+            "Num Agents",
+            "Total Time (hr)",
+            "Avg Frames per Scene",
+            "Avg Agents per Frame",
+            "Avg Scene Time (sec)",
+            "Avg Frame frequency",
+        ]
+        if len(self.agents) > 1:
+            times = self.frames[1:10]["timestamp"] - self.frames[0:9]["timestamp"]
```
oliver-scheel, line 129: that should fail when there are less than 10 frames right?

l5kit/l5kit/data/zarr_dataset.py
```diff
@@ -109,3 +113,35 @@ def open(self, mode: str = "r", cached: bool = True, cache_size_bytes: int = int
...
+            frequency = np.mean(1 / (times / 1e9))  # from nano to sec
+        else:
+            frequency = 0  # can't read the frequency
+
+        values = [
+            len(self.scenes),
+            len(self.frames),
+            len(self.agents),
+            len(self.frames) / max(frequency, 1e-9) / 3600,
```
oliver-scheel, line 138: Minor: max(..., 1e-9) is for catching div by 0? personally do not like it too much that the resulting value after the division then is a high value ... would prefer to get something like 0 - or what is the content of the zarr file then?

l5kit/l5kit/data/zarr_dataset.py
```diff
@@ -109,3 +113,35 @@ def open(self, mode: str = "r", cached: bool = True, cache_size_bytes: int = int
...
+            len(self.frames) / max(len(self.scenes), 1e-9),
+            len(self.agents) / max(len(self.frames), 1e-9),
+            len(self.frames) / max(len(self.scenes), 1e-9) / 10,
```
oliver-scheel, line 141: Doesn't 10 stem from our sampling frequency and thus needs be changed as well?

l5kit/l5kit/data/zarr_dataset.py
lucabergamini: the only cases when this fails is when there are 0 or 1 frames (mean fails). 10 is just a small random number to avoid loading a lot of data. I can put a constant there, but It has no semantical meaning so I'm not sure about it...a comment maybe?

l5kit/l5kit/data/zarr_dataset.py
lucabergamini: good catch :)

l5kit/l5kit/data/zarr_dataset.py
lucabergamini: I'll see what I can do

l5kit/l5kit/data/zarr_dataset.py
yye-lyft: In case `len(self.scenes) = 0` then `len(self.frames)` should also be 0.
To avoid divided by zero, I would simply put 1 in the denominator:
            len(self.frames) / max(len(self.scenes), 1),
            len(self.agents) / max(len(self.frames), 1),
            len(self.frames) / max(len(self.scenes), 1) / frequency,

l5kit/l5kit/data/zarr_dataset.py
yye-lyft: `if len(self.agents) > 1 and if len(self. frames) > 9:`?

Alternatively num_sample = min(len(self. frames), 9) ...

l5kit/l5kit/data/zarr_dataset.py
```diff
@@ -109,3 +113,35 @@ def open(self, mode: str = "r", cached: bool = True, cache_size_bytes: int = int
...
+            "Avg Frames per Scene",
+            "Avg Agents per Frame",
+            "Avg Scene Time (sec)",
+            "Avg Frame frequency",
+        ]
+        if len(self.agents) > 1:
+            times = self.frames[1:10]["timestamp"] - self.frames[0:9]["timestamp"]
+            frequency = np.mean(1 / (times / 1e9))  # from nano to sec
+        else:
+            frequency = 0  # can't read the frequency
```
yye-lyft, line 132: I would put 10 here by default and write a warning here

l5kit/l5kit/data/zarr_dataset.py
lucabergamini: For well designed zarr that's true, so I think we should consider that to be always true. 1e-9 or 1 doesn't really make any difference if the numerator is 0, but let's put 1

l5kit/l5kit/data/zarr_dataset.py
lucabergamini: yeah, that makes it easier also to avoid divisionbyzero below

l5kit/l5kit/data/zarr_dataset.py
lucabergamini: btw, just noticed that it should be frames and not agents

lucabergamini: address in last commit, agents was a mistake :)

PR #48, opened 2020-07-03
Lberg/speedup long scenes

lucabergamini:
Slice the minimum array in the `.zarr`. This should speed up training with long sequences

### DIFF
--- a/l5kit/l5kit/sampling/agent_sampling.py
+++ b/l5kit/l5kit/sampling/agent_sampling.py
@@ -84,8 +84,8 @@ expressed in pixels (to be changed).
     history_slice = get_history_slice(state_index, history_num_frames, history_step_size, include_current_state=True)
     future_slice = get_future_slice(state_index, future_num_frames, future_step_size)
 
-    history_frames = frames[history_slice]
-    future_frames = frames[future_slice]
+    history_frames = frames[history_slice].copy()  # copy() required if the object is a np.ndarray
+    future_frames = frames[future_slice].copy()
 
     min_agents_index = history_frames[-1]["agent_index_interval"][0]
     if len(future_frames) > 0:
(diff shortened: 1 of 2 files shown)


l5kit/l5kit/sampling/agent_sampling.py
```diff
@@ -80,16 +80,25 @@ def generate_agent_sample(
...
     #  the history slice is ordered starting from the latest frame and goes backward in time., ex. slice(100, 91, -2)
     history_slice = get_history_slice(state_index, history_num_frames, history_step_size, include_current_state=True)
-    history_frames = frames[history_slice]
-    history_agents = filter_agents_by_frames(history_frames, agents)
-
     future_slice = get_future_slice(state_index, future_num_frames, future_step_size)
-    future_frames = frames[future_slice]
+
+    history_frames = frames[history_slice].copy()  # copy() required if the object is a np.ndarray
+    future_frames = frames[future_slice].copy()
```
yye-lyft, line 88: One option to avoid copy is to pass the agent_index_offset `min_agents_index` to the function `filter_agents_by_frames`

lucabergamini: that's true. However, that would leave the indices untouched in the rest of the function. Right now those are not used, but if you stop in debug is better to have them aligned imo. But I'm absolutely available to change it if you have a strong opinion on that @yye-lyft . I don't think it would save us a lot, given that the number of picked frames  is low compared with that of the agents

yye-lyft: @lucabergamini agree, if the copy does not affect the performance I am okay with it. Yeah, better to have frame & agents aligned

lucabergamini: cpu utils seems to be very low with this update, so I think we should be fine. If anything come up in the future regarding this, we can always implement that

lucabergamini:
@yye-lyft I would expect it to be comparable with the original (v1.0.0) version, as also that one was performing the same operation. It should be faster with history turned on though.

PR #51, opened 2020-07-03
Modified Readme.

oliver-scheel:
Added short troubleshooting paragraph to install section concerning Windows machines.

### DIFF
--- a/README.md
+++ b/README.md
@@ -78,6 +78,9 @@ prediction-dataset/
 cd l5kit
 pip install -r requirements.txt
 ```
+If you run into trouble installing L5Kit on Windows, you may need to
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -78,6 +78,9 @@ prediction-dataset/
 cd l5kit
 pip install -r requirements.txt
 ```
+If you run into trouble installing L5Kit on Windows, you may need to
+- install Pytorch and torchvision manually first (select the correct version required by your system, i.e. GPU or CPU-only) then run L5Kit install (remove the packages 'torch' and 'torchvision' from ```setup.py```)
```
lucabergamini, line 82: should we add the links to the 2 lines in the setup.py?

oliver-scheel: good call!

PR #52, opened 2020-07-06
remove docs artefacts folder

lucabergamini:
Now docs is hosted in GH pages branch, so we can safely remove this one


pascal-pfeiffer:
> @lucabergamini can we remove the git history of this folder?

It's ~16 MB, which is the majority of the repo. You do have a point here, I am just not sure if it's worth the risk.

yawei-ye (approved):
Merge this PR for now, but would be nice to remove the doc git history

PR #53, opened 2020-07-06
Lberg/add deps version

lucabergamini:
Add version to some deps (especially dev). This should ensure we don't have weird behaviours in CI (like `isort` dropping `--recursive` from 5.0 )

### DIFF
--- a/l5kit/run_tests.sh
+++ b/l5kit/run_tests.sh
@@ -22,7 +22,7 @@ lint() {
 
 isort() {
   echo "import sorting.."
-  ${PYTHON_EXECUTABLE} -m isort l5kit --check-only --recursive
+  ${PYTHON_EXECUTABLE} -m isort l5kit --check-only
 }
 
 types() {
--- a/l5kit/setup.py
+++ b/l5kit/setup.py
@@ -32,7 +32,7 @@ setup(
     ],
     extras_require={
         "dev": ["pytest==5.4.3", "mypy==0.782", "setuptools", "twine", "wheel", "pytest-cov==2.10.0", "flake8==3.8.3",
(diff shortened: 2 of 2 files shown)


l5kit/run_tests.sh
```diff
@@ -22,7 +22,7 @@ lint() {
 
 isort() {
   echo "import sorting.."
-  ${PYTHON_EXECUTABLE} -m isort l5kit --check-only --recursive
+  ${PYTHON_EXECUTABLE} -m isort l5kit --check-only
```
yye-lyft, line 25: What's the reason to remove the `--recursive`?

l5kit/setup.py
```diff
@@ -31,8 +31,9 @@
         "ipywidgets"
     ],
     extras_require={
-        "dev": ["pytest", "mypy", "setuptools", "twine", "wheel", "pytest-cov", "flake8", "black", "isort",
-                "Sphinx", "sphinx-rtd-theme", "recommonmark", "pre-commit"]
+        "dev": ["pytest==5.4.3", "mypy==0.782", "setuptools", "twine", "wheel", "pytest-cov==2.10.0", "flake8==3.8.3",
+                "black==19.10b0", "isort==5.0.4", "Sphinx==3.1.1", "sphinx-rtd-theme==0.5.0", "recommonmark==0.6.0",
```
yye-lyft, line 35: I understand to specify the version of protobuf & black, but why we need to specify the version for `Sphinx`

l5kit/run_tests.sh
lucabergamini: they drop the args and made it default. It triggers an error in 5.0.3 and a warning in 5.0.4

l5kit/setup.py
lucabergamini: I think in the long run we should specify versions for everything....it avoids breaking changes as the one from isort today

PR #55, opened 2020-07-06
Lberg/refactor sat load

lucabergamini:
Start refactoring the loading function for sat image and matrix. This need to be better documented

### DIFF
--- a/l5kit/l5kit/rasterization/satellite_rasterizer.py
+++ b/l5kit/l5kit/rasterization/satellite_rasterizer.py
@@ -1,11 +1,43 @@
+import json
+import os
 from typing import List, Optional, Tuple
 
 import cv2
 import numpy as np
 
+from ..data import DataManager
 from ..geometry import rotation33_as_yaw, transform_point, world_to_image_pixels_matrix
 from .rasterizer import Rasterizer
 from .satellite_image import get_sat_image_crop_scaled
+
+
+def _load_image_and_metadata(image_key: str, data_manager: DataManager) -> Tuple[np.ndarray, dict]:
+    """Loads image from given key and its meatadata. The metadata file should be a file with the same key except for
+    having a .json extension instead.
+
+    Args:
+        image_key (str): key to the image (e.g. ``maps/my_satellite_image.png``)
+        data_manager (DataManager): DataManager used for requiring files
+
+    Raises:
+        FileNotFoundError: Image or metadata is missing or invalid
+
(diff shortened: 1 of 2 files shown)


l5kit/l5kit/rasterization/satellite_rasterizer.py
```diff
@@ -68,7 +102,7 @@ def rasterize(
         sat_translation = transform_point(np.append(world_translation, ego_translation[2]), self.map_to_sat)
 
         # Note 1: there is a negation here, unknown why this is necessary.
-        # My best guess is because Y is flipped, maybe we can do this more elegantly.
+        # Note 2 (lberg): my best guess is that cv2 rotation are clockwise in fact.
```
laky, line 105: Isn't this the same negation as the one we do for agents rasterizer etc? My explanation is that the rotation is counterclockwise and this should be negative to align the world with the ego yaw, as you are doing the inverse rotation from rotating ego by ego_yaw. As we are working in 2D, inverse rotation is the same as rotation by negative angle. 

If we do negate twice, then yes, this might be to accommodate for cv2 rotation to be clockwise, but iirc that is not the case.

lucabergamini: nope, we don't negate anymore in `BoxRasterizer` (see [here](https://github.com/lyft/l5kit/blob/eaf31c935c7389e1e4b3cf8579a0d65ba07a6fbe/l5kit/l5kit/rasterization/box_rasterizer.py#L108)). Also, cv2 rotations are counterclock in a left-system, but we're in a right one here, so we need to compensate that. That's my best explanation right now, but I'm not claiming it's the right one xD

laky: Yes we do, here: https://github.com/lyft/l5kit/blob/e8018a5c3a11b7bde1e4e99f8e7cee3b0f3b0da0/l5kit/l5kit/geometry/transform.py#L67

So I still think this should be negated if we were in right hand system looking down the negative z-axis rotating counter-clockwise. So I wouldn't blame the minus sign here on cv2 as it should be here. Not sure how using the cv2 method manifests itself, but there should be a minus somewhere if using counterclockwise rotations. If the comment with clockwise is correct, then there actually shouldn't be a minus here.

lucabergamini:
I'll close this as it has already been take over in https://github.com/lyft/l5kit/pull/64

PR #60, opened 2020-07-09
Add params to control zarr sizes when write

lucabergamini:
This enables pre-allocating arrays, which can then be indexed to write
instead of relying on append

### DIFF
--- a/l5kit/l5kit/data/zarr_dataset.py
+++ b/l5kit/l5kit/data/zarr_dataset.py
@@ -70,23 +70,26 @@ class ChunkedStateDataset:
         if ".zarr" not in self.path:
             print("zarr dataset path should end with .zarr (for now). Open will fail for this dataset!")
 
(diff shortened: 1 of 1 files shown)


l5kit/l5kit/data/zarr_dataset.py
```diff
@@ -70,23 +70,26 @@ def __init__(self, path: str, key: str = ""):
         if ".zarr" not in self.path:
             print("zarr dataset path should end with .zarr (for now). Open will fail for this dataset!")
 
-    def initialize(self, mode: str = "w") -> None:
+    def initialize(self, mode: str = "w", scenes_len: int = 0, frames_len: int = 0, agents_len: int = 0) -> None:
```
yye-lyft, line 73: Minor, maybe scene_number?
scene length sounds like how many frames are in a scene

PR #62, opened 2020-07-10
Lberg.refactor agents mask

lucabergamini:
Complete refactor of `select_agents`:
- now the mask is computed based only on `filter_agents_threshold`. You will need to run it only once and it will work for every combination of `history_num_frames` and `future_num_frames`. The boolean mask has been replaced with two uint arrays, and every agent instance now has (available_frame_past, available_frame_futures), which are used at runtime to select the agents to train or validate onto (this can still be bypassed by providing a boolean mask when creating the `AgentDataset`)
- now `select_agents` has tests
- the threshold over the movement (previously set to 3m) has been removed. @pondruska suggests we should keep parked cars as well as those waiting at crossings. This gives us almost 3x agents


l5kit/l5kit/dataset/agent.py
```diff
@@ -36,34 +40,27 @@ def load_agents_mask(self) -> np.ndarray:
...
-            agents_mask = self.dataset.root[f"agents_mask/{group_name}"]
+            agents_mask = self.dataset.root[f"agents_mask/{agent_prob}"]
         except KeyError:
             print(
-                f"cannot find {group_name} in {self.dataset.path},\n"
-                f"your cfg has loaded history_num_frames={history_num_frames}, future_num_frames={future_num_frames} "
-                f"and filter_agents_threshold={agent_prob};\n"
-                "but those values don't have a match among the agents_mask in the zarr\n"
-                "Mask will now be generated for these parameters."
+                f"cannot find the right config in {self.dataset.path},\n"
```
oliver-scheel, line 49: Minor: do we really want to throw an error for this or just a warning?

lucabergamini: it's not throwing an error actually , just printing a warning before computing the masks

l5kit/l5kit/dataset/select_agents.py
```diff
@@ -95,7 +82,7 @@ def get_valid_agents(
     agents = dataset.agents[agents_range_start:agents_range_end]
 
     agents_dict = defaultdict(list)
-    agents_selected_mask = np.zeros(len(agents), dtype=np.bool)
+    agents_mask = np.zeros((len(agents), 2), dtype=np.int)
```
oliver-scheel, line 85: Can we have a comment here like in the PR description telling what agents_mask means? (available_frame_past, available_frame_futures).
In general, I have to say, I did not find this so easy to follow.

l5kit/l5kit/dataset/select_agents.py
```diff
@@ -112,82 +99,64 @@ def get_valid_agents(
...
                 continue
 
             # ==== COUPLE-WISE FILTERS
             if len(agents_dict[agent["track_id"]]) > 1:
                 p_frame_idx, p_global_agent_idx, p_agent = agents_dict[agent["track_id"]][-2]  # get prev element
-                frame_lost = get_missing_frame_num(agents_dict[agent["track_id"]][:-1], agents_selected_mask)
 
                 if not in_consecutive_frame(frame_idx, p_frame_idx):
-                    report["reject_th_hole"] += frame_lost
+                    update_mask(agents_mask, agents_dict[agent["track_id"]][:-1])
```
oliver-scheel, line 118: Could think about shortening this, as there are nearly 3 duplicate lines for each case.

lucabergamini: because we need to fill the report we won't save any lines sadly :(

l5kit/l5kit/dataset/agent.py
yye-lyft: I think we should swich to logger to have proper warning at some point

PR #68, opened 2020-07-15
get_combined_scenes is now not default

lucabergamini:
`get_combined_scenes` is now not called automatically in `dataloader_builder`. This fix an issue with the first run of AgentDataset when the mask is computed because:
- the ChunkedStateDataset is created and scenes aggregated
- the object is stored in `self.datataset` for the `AgentDataset` by super call. This also set `cumulative_sizes` using scenes (aggregated)
- mask is computed (require re-opening the dataset which is performed without calling `get_combined_scenes`) and **root updated** (frames, scenes and agents reloaded)
- `cumulative_sizes_agents` is set

Clearly, the two cumulative sizes don't match. If the script is run again, `cumulative_sizes_agents` is created using the aggregated scenes instead and everything works.

Note: this was having an impact on the `agents_mask` (not a dramatic one actually)

### DIFF
--- a/l5kit/l5kit/dataset/dataloader_builder.py
+++ b/l5kit/l5kit/dataset/dataloader_builder.py
@@ -15,7 +15,7 @@ def build_dataloader(
     dataset_class: Callable,
     rasterizer: Rasterizer,
     perturbation: Optional[Perturbation] = None,
-    combine_scene: bool = False,
+    combine_scenes: bool = False,
 ) -> DataLoader:
     """
     Function to build a dataloader from a dataset of dataset_class. Note we have to pass rasterizer and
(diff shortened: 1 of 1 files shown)


l5kit/l5kit/dataset/dataloader_builder.py
```diff
@@ -15,6 +15,7 @@ def build_dataloader(
     dataset_class: Callable,
     rasterizer: Rasterizer,
     perturbation: Optional[Perturbation] = None,
+    combine_scene: bool = False,
```
yye-lyft, line 18: nit: `combine_scenes`

l5kit/l5kit/dataset/dataloader_builder.py
```diff
@@ -27,6 +28,7 @@ def build_dataloader(
         dataset_class (Callable): a class object (EgoDataset or AgentDataset currently) to build the dataset
         rasterizer (Rasterizer): the rasterizer for the dataset
         perturbation (Optional[Perturbation]): an optional perturbation object
+        combine_scene (float): if to combine scenes that follow up each other perfectly
```
yye-lyft, line 31: boolean

PR #70, opened 2020-07-16
replace path with dataset for select_agents, avoid calling open again in AgentDataset

lucabergamini:
Replace first argument of select_agents (Path) with the `ChunkedStateDataset` itself. This avoid performing an `open()` call when `AgentDataset` builds the mask at runtime.

Note this is slightly related to #68. When both will be merged the behaviour will be back to normal (plus select_agents will be more accurate)

### DIFF
--- a/l5kit/l5kit/dataset/agent.py
+++ b/l5kit/l5kit/dataset/agent.py
@@ -1,7 +1,9 @@
 import bisect
+from pathlib import Path
 from typing import Optional
 
 import numpy as np
+from zarr import convenience
 
 from ..data import ChunkedStateDataset
 from ..kinematic import Perturbation
@@ -61,8 +63,8 @@ class AgentDataset(EgoDataset):
                 th_movement=TH_MOVEMENT,
                 th_distance_av=TH_DISTANCE_AV,
             )
-            self.dataset.open()  # ensure root is updated, reload all arrays TODO lberg: avoid reload
-            agents_mask = self.dataset.root[f"agents_mask/{group_name}"]
+            array_path = Path(self.dataset.path) / f"agents_mask/{group_name}"
+            agents_mask = convenience.load(str(array_path))  # note (lberg): this doesn't update root
 
         return agents_mask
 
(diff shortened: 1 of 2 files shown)


l5kit/l5kit/dataset/agent.py
```diff
@@ -52,17 +52,17 @@ def load_agents_mask(self) -> np.ndarray:
...
                 history_num_frames,
                 future_num_frames,
                 th_yaw_degree=TH_YAW_DEGREE,
                 th_extent_ratio=TH_EXTENT_RATIO,
                 th_movement=TH_MOVEMENT,
                 th_distance_av=TH_DISTANCE_AV,
-                num_workers=8,
-            )  # TODO maybe set in env var?
-            self.dataset.open()  # ensure root is updated
+                num_workers=8,  # TODO maybe set in env var?
```
yye-lyft, line 63: How about remove it here. Just use the `cpu_count` that `multiprocessing package` gives you for the paralleisation in `select_agents.py`?
`from multiprocessing import cpu_count, Pool`
...
`with Pool(cpu_count()) as pool:`

l5kit/l5kit/dataset/select_agents.py
```diff
@@ -266,8 +262,11 @@ def select_agents(
     args = parser.parse_args()
 
     for input_folder in args.input_folder:
```
yye-lyft, line 264: `args.input_folders`?

l5kit/l5kit/dataset/agent.py
```diff
@@ -52,17 +52,17 @@ def load_agents_mask(self) -> np.ndarray:
...
+            )
+            self.dataset.open()  # ensure root is updated, reload all arrays TODO lberg: avoid reload
```
yye-lyft, line 65: why we need to call `open` again?

lucabergamini: to update the root dict basically (otherwise you get a KeyError again). Let me see if I can find a better way

yye-lyft: it can be in a subsequent PR, I will approve it for now

PR #72, opened 2020-07-16
Lberg/fix arrow visualisation

lucabergamini:
Fix an issue affecting arrow visualisation. Refactor code to use cv2 rotation matrix.
NOTE: In the following 2 pictures the arrows have been magnified and their number reduced

BEFORE
Screenshot 2020-07-16 at 13 31 26

AFTER:
Screenshot 2020-07-16 at 13 32 10

NOTE: I've increased the tip length, now arrows look like:
Screenshot 2020-07-16 at 13 36 24

Are we fine with that?

### DIFF
--- a/l5kit/l5kit/visualization/utils.py
+++ b/l5kit/l5kit/visualization/utils.py
@@ -9,8 +9,9 @@ PREDICTED_POINTS_COLOR = (0, 255, 255)
 TARGET_POINTS_COLOR = (255, 0, 255)
 REFERENCE_TRAJECTORY_POINT_COLOR = (255, 255, 0)
 #  Arrows represent position + orientation.
-ARROW_LENGTH_IN_PIXELS = 4
+ARROW_LENGTH_IN_PIXELS = 2
 ARROW_THICKNESS_IN_PIXELS = 1
+ARROW_TIP_LENGTH = 1.8
 
 
 def draw_arrowed_line(on_image: np.ndarray, position: np.ndarray, yaw: float, rgb_color: Tuple[int, int, int]) -> None:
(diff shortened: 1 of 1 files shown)


l5kit/l5kit/visualization/utils.py
```diff
@@ -3,14 +3,15 @@
 import cv2
 import numpy as np
 
-from l5kit.geometry import transform_points
+from l5kit.geometry import transform_point, transform_points
```
laky, line 6: Very nit: do we really need 2 methods for this in the geometry module? Might be neater to have just the transform_points and send in array of 1 point. I'm just worried if you make a typo in 1 letter, you may get a very different behaviour I assume.

laky, line 6: Very nit: do we really need 2 methods for this in the geometry module? Might be neater to have just the transform_points and send in array of 1 point. I'm just worried if you make a typo in 1 letter, you may get a very different behaviour I assume.

l5kit/l5kit/visualization/utils.py
```diff
@@ -25,11 +26,18 @@ def draw_arrowed_line(on_image: np.ndarray, position: np.ndarray, yaw: float, rg
...
+    end_pixel = start_pixel + transform_point(np.asarray([ARROW_LENGTH_IN_PIXELS, 0]), rot)
+
     cv2.arrowedLine(
-        on_image, tuple(start_pixel), tuple(end_pixel), rgb_color, thickness=ARROW_THICKNESS_IN_PIXELS, tipLength=0.4
+        on_image,
+        tuple(start_pixel.astype(np.int32)),
+        tuple(end_pixel.astype(np.int32)),
+        rgb_color,
+        thickness=ARROW_THICKNESS_IN_PIXELS,
+        tipLength=ARROW_TIP_LENGTH,
```
laky, line 40: Nit: `ARROW_TIP_LENGTH_IN_PIXELS`

laky (dismissed):
LGTM as a fix to a broken visualisation.

Let's think separately whether there might be a better way to visualise this perhaps.

l5kit/l5kit/visualization/utils.py
lucabergamini: I think I asked the same thing to Guido once xD

yye-lyft (approved):

PR #73, opened 2020-07-17
Lberg/refactor test suite

lucabergamini:
Refactor of L5Kit tests:

- drop `unittest`;
- move all artefacts into a folder `artefacts` so that we can simulate data loading. Add metas;
- remove `agents_mask` (now computed at runtime)
- remove all dependency to files in l5kit (config mainly)
- remove all hard-coded matrices
- remove all mock functions to get rasterizers (replaced with functions from l5kit finally)

This should make tests much easier to maintain and develop.

PS
Sorry for the very long PR folks, but breaking it down into a lot of small ones was not trivial because of data dependencies. hopefully, it will be the only one this long in a while :)


l5kit/l5kit/tests/rasterization/box_rasterizer_test.py
```diff
@@ -27,48 +28,67 @@ def test_draw_boxes() -> None:
...
+    assert out[..., 0].sum() == 0
+
+    cfg["raster_params"]["filter_agents_threshold"] = 0.0
+    rasterizer = build_rasterizer(cfg, dm)
+
+    out = rasterizer.rasterize(hist_data[0][:1], hist_data[1][:1])
+    assert out[..., 0].sum() > 0
+
+
+def test_agent_ego(hist_data: tuple) -> None:
```
yye-lyft, line 70: `test_agent_as_ego` or something more telling?

PR #76, opened 2020-07-21
Change agent_mask check

lucabergamini:
Instead of checking the group for the `agents_mask`, check directly the full path.
This avoids calling multiple `select_agents` if the root is not updated (can be updated only by calling `open()` again, but that replace arrays).

### DIFF
--- a/l5kit/l5kit/dataset/select_agents.py
+++ b/l5kit/l5kit/dataset/select_agents.py
@@ -6,6 +6,7 @@ from collections import Counter, defaultdict
 from functools import partial
 from multiprocessing import Pool, cpu_count
 from pathlib import Path
+from tempfile import gettempdir
 from typing import Tuple
 from uuid import uuid4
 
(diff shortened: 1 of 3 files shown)


l5kit/l5kit/dataset/select_agents.py
```diff
@@ -181,7 +180,7 @@ def select_agents(
         pass  # group is already there
 
     agents_mask = zarr.open_array(
-        str(Path(zarr_dataset.path) / "agents_mask" / output_group),
+        str(Path(zarr_dataset.path) / "agents_mask" / f"{th_agent_prob}"),
```
yye-lyft, line 183: `f"{zarr_dataset.path}/agents_mask/{th_agent_prob}"` looks simpler

yye-lyft: Since the filename has been used at two places, LINE161, LINE183, we can define a variable for it

lucabergamini: ok for the variable, but I will keep it in Path as ` f"{zarr_dataset.path}/agents_mask/{th_agent_prob}"` is platform specific.

PR #81, opened 2020-07-23
Add last debug rasterizer, create pose_to_ecef fun

lucabergamini:
Now `pose_to_ecef` hardcoded matrix lives in its own function (still TODO
to remove it when meta is in dataset)
Also add the last debug_rasterizer (semantic only)


l5kit/l5kit/rasterization/rasterizer_builder.py
```diff
@@ -148,21 +148,19 @@ def build_rasterizer(cfg: dict, data_manager: DataManager) -> Rasterizer:
...
-                    [-3.05311332e-16, 7.94103464e-01, 6.07782600e-01, 3.85516476e06],
-                    [0.00000000e00, 0.00000000e00, 0.00000000e00, 1.00000000e00],
-                ],
-                dtype=np.float64,
-            )
+            pose_to_ecef = get_hardcoded_pose_to_ecef()
 
         map_to_sat = np.matmul(ecef_to_sat, pose_to_ecef)
         return SatelliteRasterizer(raster_size, pixel_size, ego_center, sat_image, map_to_sat)
+    elif map_type == "semantic_debug":
```
oliver-scheel, line 155: What do you think of managing this in the py_semantic branch, as only the line containing the class instantiation is different?
Note scope of the PR, but also holds for satellite map.

lucabergamini: yeah I don't see why not :) I'll do it for both here

PR #85, opened 2020-07-29
Refactor agents_mask

lucabergamini:
Add the following
- `AgentDataset` now checks using fixed values by default (and warns otherwise)
- default for history is 10 (11 with present)
- default for future is 1 (at least a valid step)

- [x] update `AgentDataset`
- [x] write future availability in csv
- [x] support availability in metrics (metrics are also changing right now)
- [x] support availability in notebook forward
- [x] support availability in notebook csv write


lucabergamini:
> Question: What does the number 12 mean in line 12 in compute_mse_loss_test.py?

I've added a param for that. We could use `kargs` also, but then we would have to specify also `filter_agents_threshold` which will break all the lines -.-

P.S. you can add comments on specific lines



> Also, I'd suggest adding some more comments to the code in the Jupyter notebook. The code might be hard to follow if not explained exactly what you are doing.

Can you maybe take that over?



> Comment: In almost all functions and files we don't have a docstring or a simple comment that explains the function but that can maybe be added later if this is merged into master.

It would be great to have a list of all the functions missing the docstring actually...any chance you can take that over too :) ?

l5kit/l5kit/evaluation/extract_ground_truth.py
```diff
@@ -21,25 +19,22 @@ def export_zarr_to_ground_truth_csv(
...
         history_step_size (int): Steps to take between frames, can be used to subsample history frames.
         future_step_size (int): Steps to take between targets into the future.
-        based on their probability of being a relevant agent.
     """
 
     assert future_step_size == history_step_size == 1, "still not handled in select_agents"
     cfg = {
         "raster_params": {
-            "pixel_size": pixel_size,
+            "pixel_size": np.asarray((0.25, 0.25)),
```
laky, line 31: Why is this changed?

l5kit/l5kit/dataset/agent.py
```diff
@@ -11,6 +11,9 @@
 from .ego import EgoDataset
 from .select_agents import TH_DISTANCE_AV, TH_EXTENT_RATIO, TH_YAW_DEGREE, select_agents
 
+# WARNING: changing this value impact the number of instances selected for both train and inference!
+MIN_FRAME_HISTORY = 10  # minimum number of frames an agents must have in the past to be picked
```
laky, line 15: Is it possible to have the MIN_FRAME_FUTURE as well? Sorry for flip-floping here, but it might be good if we can reproduce the current behaviour in the future if needed. 

Could you extract the other "magic constants" like the distance threshold and the box extent and yaw changes?

It would be also worth considering putting them into config.yml. Then we can change them without having to bake new versions of master. We can also then have a separate setting for training config and for evaluation config (the latter should be kept constant and submitted to the repo).

examples/agent_motion_prediction/agent_motion_prediction.ipynb
```diff
@@ -99,10 +98,13 @@
...
     "def forward(data, model, device, criterion):\n",
     "    inputs = data[\"image\"].to(device)\n",
-    "    targets = data[\"target_positions\"].to(device).reshape(len(data[\"target_positions\"]), -1)\n",
+    "    target_availabilities = data[\"target_availabilities\"].to(device)\n",
+    "    targets = data[\"target_positions\"].to(device)\n",
     "    # Forward pass\n",
-    "    outputs = model(inputs)\n",
+    "    outputs = model(inputs).reshape(targets.shape)\n",
     "    loss = criterion(outputs, targets)\n",
+    "    # not all the output are valid, but we can filter them out from the loss using availabilities\n",
```
laky, line 106:
```suggestion
    "    # not all the output steps are valid, but we can filter them out from the loss using availabilities\n",
```

l5kit/l5kit/dataset/agent.py
lucabergamini: I can add the `MIN_FRAME_FUTURE`. Should I set it to default 0 then?
For the magic constants I'm nor really sure though. They are defined in the script that use them, where would you put them? 

I'm not sure about that...it depends on what we think these values are. If they are something that should never be touched, then config is not really the right place imo.

l5kit/l5kit/evaluation/extract_ground_truth.py
lucabergamini: That param doesn't change prediction values are they are in meters now, so I've removed it from args of this function

examples/agent_motion_prediction/agent_motion_prediction.ipynb
```diff
@@ -72,7 +72,7 @@
     "    # load pre-trained Conv2D model\n",
     "    model = resnet50(pretrained=True)\n",
     "\n",
-    "    # change input size\n",
+    "    # change input size to address for the rasterizer output\n",
```
oliver-scheel, line 75: to match the rasterizer's output?

examples/agent_motion_prediction/agent_motion_prediction.ipynb
```diff
@@ -83,8 +83,7 @@
     "        padding=model.conv1.padding,\n",
     "        bias=False,\n",
     "    )\n",
-    "    # change output size\n",
-    "    # X, Y  * number of future states\n",
+    "    # change output size to (X, Y)  * number of future states\n",
```
oliver-scheel, line 86: remove space before *

l5kit/l5kit/dataset/agent.py
```diff
@@ -11,6 +11,10 @@
 from .ego import EgoDataset
 from .select_agents import TH_DISTANCE_AV, TH_EXTENT_RATIO, TH_YAW_DEGREE, select_agents
 
+# WARNING: changing these values impact the number of instances selected for both train and inference!
+MIN_FRAME_HISTORY = 10  # minimum number of frames an agents must have in the past to be picked
+MIN_FRAME_FUTURE = 0  # minimum number of frames an agents must have in the future to be picked
```
oliver-scheel, line 16: MIN_FRAME_HISTORY and _FUTURE are +1 right, ie MIN_FRAME_HISTORY = 10 means 10 past frames + current right? 
Does it make sense to set _FUTURE to 0 then, or should it not be 1? (As agents with no future frame are never used.)

l5kit/l5kit/evaluation/extract_ground_truth.py
```diff
@@ -21,25 +19,22 @@ def export_zarr_to_ground_truth_csv(
...
             "raster_size": (100, 100),
             "filter_agents_threshold": filter_agents_threshold,
             "ego_center": np.asarray((0.5, 0.5)),
         },
         "model_params": {
-            "history_num_frames": history_num_frames,
+            "history_num_frames": 0,
```
oliver-scheel, line 37: Do we actually need to create a cfg with some unused parameters or could we just pass them to the constructur as is?

l5kit/l5kit/dataset/agent.py
lucabergamini: only history is +1, so 10 means present plus 10 frames.
future 0 means the future is not considered (e.g. can all be invalid). I also think 1 is a better value, let's put 1 @oliver-scheel

l5kit/l5kit/evaluation/extract_ground_truth.py
lucabergamini: trouble here is that the partial inside read the configuration, so we need to pass a kind-of valid config..I agree it sucks though xD

l5kit/l5kit/evaluation/extract_losses.py
```diff
@@ -18,22 +18,29 @@ def extract_csv(csv_path: str) -> list:
     ground_truth_rows = extract_csv(ground_truth_path)[1:]
     inference_rows = extract_csv(inference_output_path)[1:]
 
+    assert len(ground_truth_rows[0]) == len(inference_rows[0]), "number of fields do not match"
+
+    future_num_frames = int(ground_truth_rows[0][2])  # read future_num_frames TODO: make this less mysterious
```
oliver-scheel, line 23: Also not a huge fan of using these indices etc I have to say - maybe I'm not seeing it, but any reason why not work with column names (e.g. using np.savetxt for saving a csv or pandas should work)?

oliver-scheel: Also connecting to e.g. the parse_values function in general, but maybe scope of a different PR.

l5kit/l5kit/evaluation/extract_losses.py
```diff
@@ -62,10 +69,18 @@ def compute_mse(A: np.ndarray, B: np.ndarray) -> np.ndarray:
...
+        gt_coords = ground_truth_value["coords"]
+        pred_coords = inference[key]["coords"]
+
+        errors.append(compute_mse(gt_coords, pred_coords))
+        gt_avails.append(ground_truth_value["avail"])
 
-    return np.array(errors).mean(axis=0)  # reduce samples, keep steps
+    # use numpy masked package to get a mean where unavailable elements are masked out
+    mask = ~np.array(gt_avails).astype(np.bool)  # note the ~ here, masked seems to assume 0 means keep
+    errors = np.ma.masked_array(np.array(errors), mask=mask)
```
oliver-scheel, line 82: Why use np masked instead of simply errors[mask]?

l5kit/l5kit/evaluation/extract_losses.py
lucabergamini: This will be completely re-written @oliver-scheel . It's a very old code indeed :)

l5kit/l5kit/evaluation/extract_losses.py
lucabergamini: because the result of that cannot be reshaped (it's a flatten array) to keep the `future_step` axis, and we want to provide a value for each timestep

l5kit/l5kit/dataset/agent.py
laky: Fair, let's keep it this way. I can see value for changing this for training, but now that I think about it again, we can use the availability masks in GT to do this filtering, right?

lucabergamini: the mask can always be passed by the caller, so I would say yes

PR #92, opened 2020-08-05
Add csv utils function

lucabergamini:
add functions for:
- writing GT csv
- reading GT csv
- writing pred csv
- reading pred-csv

The reason we write them with different function is because GT has
availability and it's single mode, while pred has multi-mode with
confidences

### DIFF
--- a/l5kit/l5kit/evaluation/csv_utils.py
+++ b/l5kit/l5kit/evaluation/csv_utils.py
@@ -6,6 +6,23 @@ import numpy as np
 
 MAX_MODES = 3
 
+"""
+These utilities write and read csv with ground-truth and prediction data.
+Both share the first two field (timestamp and track_id) which are used to identify a single record.
+
+GT:
+Single mode future prediction with availabilities (either 1->available or 0->unavailable).
+Header fields have these meanings:
+timestamp, track_id, avail_time_0, avail_time_1, ..., coord_x_time_0, coord_y_time_0, ...
+
+PRED:
+Multi mode future prediction with confidence score (one value per mode up to MAX_MODES, sum to 1).
+Header fields have these meanings:
+timestamp, track_id, conf_mode_0, conf_mode_1, ..., coord_x_time_0_mode_0, coord_y_time_0_mode_0, ...,
+coord_x_time_0_mode_1, coord_y_time_0_mode_1, ...
+
+"""
+
 
 def _generate_coords_keys(future_len: int, mode_index: int = 0) -> List[str]:
     """
--- a/l5kit/l5kit/evaluation/csv_utils.py
+++ b/l5kit/l5kit/evaluation/csv_utils.py
@@ -8,7 +8,7 @@ MAX_MODES = 3
 
 """
 These utilities write and read csv with ground-truth and prediction data.
-Both share the first two field (timestamp and track_id) which are used to identify a single record.
+Both share the first two fields (timestamp and track_id) which are used to identify a single record.
 
 GT:
 Single mode future prediction with availabilities (either 1->available or 0->unavailable).
@@ -30,7 +30,7 @@ def _generate_coords_keys(future_len: int, mode_index: int = 0) -> List[str]:
     Two keys for each mode and future step.
 
     Args:
-        future_len (int): how many prediction in the future data has
+        future_len (int): how many prediction the data has in the future
         mode_index (int): what mode are we reading/writing
 
     Returns:
@@ -91,9 +91,7 @@ def write_gt_csv(
 
     # create and write HEADER
     # order is (timestamp,track_id,availabilities,coords)
-    fieldnames = ["timestamp", "track_id"]
-    fieldnames.extend(avails_keys)
-    fieldnames.extend(coords_keys)
+    fieldnames = ["timestamp", "track_id"] + avails_keys + coords_keys
     writer = csv.DictWriter(open(csv_path, "w"), fieldnames)
     writer.writeheader()
 
@@ -109,7 +107,7 @@ def write_gt_csv(
 
 def read_gt_csv(csv_path: str) -> Iterator[dict]:
     """
-    Generator function that returns a line at the time from the csv file as a dict
+    Generator function that returns a line at a time from the csv file as a dict
 
     Args:
         csv_path (str): path of the csv to read
@@ -150,8 +148,8 @@ def write_pred_csv(
     """
     Encode the ground truth into a csv file. Coords can have an additional axis for multi-mode.
     We handle up to MAX_MODES modes.
-    For the uni-modal case, coords should not have the additional axis and confs should be set to None.
-    In this case, a single mode with confidence 1 will be written.
+    For the uni-modal case (i.e. all predictions have just a single mode), coords should not have the additional axis
+    and confs should be set to None. In this case, a single mode with confidence 1 will be written.
 
     Args:
         csv_path (str): path to the csv to write
@@ -188,8 +186,7 @@ def write_pred_csv(
 
     # create and write HEADER
     # order is (timestamp,track_id,confidences,coords)
-    fieldnames = ["timestamp", "track_id"]
-    fieldnames.extend(confs_keys)  # all confidences before coordinates
+    fieldnames = ["timestamp", "track_id"] + confs_keys  # all confidences before coordinates
     for coords_labels in coords_keys_list:
         fieldnames.extend(coords_labels)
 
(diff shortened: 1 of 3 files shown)


l5kit/l5kit/evaluation/csv_utils.py
```diff
@@ -0,0 +1,246 @@
...
+from itertools import chain
+from typing import Iterator, List, Optional
+
+import numpy as np
+
+MAX_MODES = 3
+
+"""
+These utilities write and read csv with ground-truth and prediction data.
+Both share the first two field (timestamp and track_id) which are used to identify a single record.
```
stefaniespeichert, line 11: 'First two fieldS' (missing s)

l5kit/l5kit/evaluation/csv_utils.py
```diff
@@ -0,0 +1,246 @@
...
+"""
+
+
+def _generate_coords_keys(future_len: int, mode_index: int = 0) -> List[str]:
+    """
+    Generate keys like coord_x00, coord_y00... that can be used to get or set value in CSV.
+    Two keys for each mode and future step.
+
+    Args:
+        future_len (int): how many prediction in the future data has
```
stefaniespeichert, line 33: 'how many prediction in the future data has'
not exactly clear what you mean here. Maybe say:  how many predictions the data has in the future ?

l5kit/l5kit/evaluation/csv_utils.py
```diff
@@ -0,0 +1,246 @@
...
+    fieldnames.extend(avails_keys)
+    fieldnames.extend(coords_keys)
+    writer = csv.DictWriter(open(csv_path, "w"), fieldnames)
+    writer.writeheader()
+
+    for timestamp, track_id, coord, avail in zip(timestamps, track_ids, coords, avails):
+        # writing using dict relieves us from respecting the order
+        line = {"timestamp": timestamp, "track_id": track_id}
+        line.update({key: ava for key, ava in zip(avails_keys, avail)})
+        # TODO limit digits in coords?
```
stefaniespeichert, line 104: Is it ok to have that todo still in?

l5kit/l5kit/evaluation/csv_utils.py
```diff
@@ -0,0 +1,246 @@
...
+        line.update({key: cor for key, cor in zip(coords_keys, coord.reshape(-1))})
+
+        writer.writerow(line)
+
+
+def read_gt_csv(csv_path: str) -> Iterator[dict]:
+    """
+    Generator function that returns a line at the time from the csv file as a dict
```
stefaniespeichert, line 112: 'a line at A time' (since you're iterating you should use 'a' instead of the.)

l5kit/l5kit/evaluation/csv_utils.py
```diff
@@ -0,0 +1,246 @@
...
+    assert timestamps.shape == track_ids.shape == (num_example,)
+    assert confs is not None and confs.shape == (num_example, num_modes)
+    assert num_modes <= MAX_MODES
+
+    # generate always a fixed size json for MAX_MODES by padding the arrays with zeros
+    coords_padded = np.zeros((num_example, MAX_MODES, future_len, num_coords), dtype=coords.dtype)
+    coords_padded[:, :num_modes] = coords
+    confs_padded = np.zeros((num_example, MAX_MODES), dtype=confs.dtype)
+    confs_padded[:, :num_modes] = confs
+    # TODO check sum to 1
```
stefaniespeichert, line 185: Leave TODO in here?

l5kit/l5kit/tests/evaluation/csv_utils_test.py
```diff
@@ -0,0 +1,124 @@
...
+
+    # read and check values
+    for idx, el in enumerate(read_pred_csv(dump_path)):
+        assert int(el["track_id"]) == track_ids[idx]
+        assert int(el["timestamp"]) == timestamps[idx]
+        assert np.allclose(el["coords"], coords[idx], atol=1e-4)
+        assert np.allclose(el["conf"], confs[idx], atol=1e-4)
+
+
+def test_e2e_single_pred_csv(tmpdir: Path) -> None:
```
stefaniespeichert, line 106: If these tests should be used by the kaggle participants, definitely add some documentation to each function otherwise you'll be swamped later!

l5kit/l5kit/evaluation/csv_utils.py
lucabergamini: I need to talk with @pondruska about this

l5kit/l5kit/evaluation/csv_utils.py
lucabergamini: yes for now, I have a branch that fixes this (but remind it to me in a week xD)

l5kit/l5kit/tests/evaluation/csv_utils_test.py
lucabergamini: no it shouldn’t :) tests documentation is just not there at the moment...we'll have to fix it in the future

l5kit/l5kit/evaluation/csv_utils.py
```diff
@@ -0,0 +1,246 @@
...
+MAX_MODES = 3
+
+"""
+These utilities write and read csv with ground-truth and prediction data.
+Both share the first two field (timestamp and track_id) which are used to identify a single record.
+
+GT:
+Single mode future prediction with availabilities (either 1->available or 0->unavailable).
+Header fields have these meanings:
+timestamp, track_id, avail_time_0, avail_time_1, ..., coord_x_time_0, coord_y_time_0, ...
```
oliver-scheel, line 16: How about abbreviating this a bit by mentioning there are N timesteps and then writing availability_0, x_0, ... availalbility_N, x_N, y_N

l5kit/l5kit/evaluation/csv_utils.py
```diff
@@ -0,0 +1,246 @@
...
+    assert timestamps.shape == track_ids.shape == (num_example,)
+    assert avails.shape == (num_example, future_len)
+
+    coords_keys = _generate_coords_keys(future_len)
+    avails_keys = _generate_avails_keys(future_len)
+
+    # create and write HEADER
+    # order is (timestamp,track_id,availabilities,coords)
+    fieldnames = ["timestamp", "track_id"]
+    fieldnames.extend(avails_keys)
```
oliver-scheel, line 95: Minor: Do you prefer this over ["timestamp", "track_id"] + avails_keys .... ?

l5kit/l5kit/evaluation/csv_utils.py
oliver-scheel: Do we need this check - which branch is it @lucabergamini ?
Should continue discussion there then, but my thoughts for now: What error tolerance do we allow?
Should our metric not penalize e.g. only confident predictions, and we leave it to the users how they would like to handle confidence?

l5kit/l5kit/evaluation/csv_utils.py
oliver-scheel: Also here: what is the reason for this? limit digs = actual digits or cap large numbers?

l5kit/l5kit/evaluation/csv_utils.py
```diff
@@ -0,0 +1,246 @@
...
+    csv_path: str,
+    timestamps: np.ndarray,
+    track_ids: np.ndarray,
+    coords: np.ndarray,
+    confs: Optional[np.ndarray] = None,
+) -> None:
+    """
+    Encode the ground truth into a csv file. Coords can have an additional axis for multi-mode.
+    We handle up to MAX_MODES modes.
+    For the uni-modal case, coords should not have the additional axis and confs should be set to None.
```
oliver-scheel, line 153: Minor: "For the uni-modal case' sounds a bit like there can be multi-modal models which only output a single prediction for certain umabigious cases, and in this case should fall back to outputting the uni-modal format (whereas, we just distinguish between models which always output uni-modal predicitions and models which always output multi-modal predictions). With a good explanation of this e.g. in the competition this should be ok though.

l5kit/l5kit/evaluation/csv_utils.py
lucabergamini: The first one, we may want to write up to 3 decimal (millimetres) or something like that. It would really help compression and general size :)

l5kit/l5kit/evaluation/csv_utils.py
lucabergamini: not really, I'm fine with replacing it :)

l5kit/l5kit/evaluation/csv_utils.py
lucabergamini: I've added a comment to ensure it's clear that every prediction must be uni-modal

lucabergamini: TBH you could have a mixed prediction set where some predictions have just a single mode with confidence set to 1. In the end, uni-modal is just a special case of multi :). The unimodal option is to avoid to specify confidences at all really.

You're right, we should clarify it better in the competiion

Issue #94, opened 2020-08-06
How can I get history data from dataloader (example/agent_motion_prediction.ipynb)


hayoung-kim:
Hello, I wonder how can I get history position and yaw of the vehicles from `dataloader`. 
In `agent_motion_prediction` example, the number of the history frames can be set at `agent_motion_config.yaml` file. 
However, when I get data from `dataloader` built from `build_dataloader` function, I cannot find history data.

Here is a code:
```
>>> tr_it = iter(train_dataloader)
>>> data = next(tr_it)

>>> print(data.keys())
dict_keys(['image', 'target_positions', 'target_yaws', 
  'target_availabilities', 'world_to_image', 'track_id', 
  'timestamp', 'centroid', 'yaw', 'extent'])
```

and here are configurations (`agent_motion_config.yaml`)
```
###################
## Model options
model_params:
  model_architecture: "resnet50"

  history_num_frames: 3
  history_step_size: 1
  history_delta_time: 0.1

  future_num_frames: 50
  future_step_size: 1
  future_delta_time: 0.1
```

How can I get history data?
Thanks!

lucabergamini:
Hi @hayoung-kim , we don't use history values explicitly right now (they are consumed by the rasterisers) so, differently from future values which are requested for a standard training, we don't export them. 

However, it should be fairly easy to support this by calling `_create_targets_for_deep_prediction` in `agent_sampling.py` with history information. 
If you start a PR to support it, I'm happy to review it :)

lucabergamini:
#96 has now been merged in master and will be included in `1.0.6` release

PR #96, opened 2020-08-06
Add history positions to `agent_sampling`

joseluisvaz:
I was experimenting with the l5kit datasets and I needed the (x, y, yaw) historical information. I saw that other people were needing it (#94) so I created a PR. 

There are still some questions:
 * It is unclear if history data should be sent backwards in time or if it should be reverted.
 * Should the current frame be included in `history_<>` or should it be skipped since there is already 'centroid' and `yaw`.

### DIFF
--- a/l5kit/l5kit/sampling/agent_sampling.py
+++ b/l5kit/l5kit/sampling/agent_sampling.py
@@ -138,11 +138,19 @@ to train models that can recover from slight divergence from training set data
         future_num_frames, future_frames, selected_track_id, future_agents, agent_centroid[:2], agent_yaw,
     )
 
+    # history_num_frames + 1 because it also counter the current frame
+    history_coords_offset, history_yaws_offset, history_availability = _create_targets_for_deep_prediction(
+        history_num_frames + 1, history_frames, selected_track_id, history_agents, agent_centroid[:2], agent_yaw,
+    )
+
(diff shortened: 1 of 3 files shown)


lucabergamini:
HI @joseluisvaz , 
amazing contribution, I will take a look asap!


> * It is unclear if history data should be sent backwards in time or if it should be reverted.

I would stick to the rasterizer notation for this (so backwards)



> * Should the current frame be included in `history_<>` or should it be skipped since there is already 'centroid' and `yaw`.

including it would mean a (0,0,0) in the array as the values are displacements. Not very useful, but it would again follow the notation of the rasteriser so I would go for including it.

What do you think about it?

joseluisvaz:
Hello @lucabergamini! 

I think it does make sense to follow the notation of the rasterizers to avoid confusion.

l5kit/l5kit/sampling/agent_sampling.py
```diff
@@ -138,11 +138,19 @@ def generate_agent_sample(
         future_num_frames, future_frames, selected_track_id, future_agents, agent_centroid[:2], agent_yaw,
     )
 
+    # history_num_frames + 1 because it also counter the current frame
```
lucabergamini, line 141: "because it also includes..."

PR #97, opened 2020-08-07
refactor and add util

lucabergamini:
Add a util to split `zarr` and refactor concat to use indexing instead of append

Why indexing when zarr provides the handy `concat()` method?

concat is fine for short operations (1k zarrs), but performance decreases linearly with time (this could be a performance bug).

On 100k input zarrs, indexing doesn't suffer from performance degradation, while at 25% concat has already dropped from 50 it/s to 24 it/s.
The downside is that indexing is way more error prone.


l5kit/l5kit/data/zarr_utils.py
```diff
@@ -1,66 +1,223 @@
...
-        print(f"output tl_faces size: {len(output_dataset.tl_faces)}")
+        _append_zarr_subset(input_dataset, output_dataset, 0, len(input_dataset.scenes), cur_num_els)
+        cur_num_els += Counter(num_els_valid_zarrs[idx])
+
+
+def zarr_split(input_zarr: str, output_path: str, split_infos: List[dict]) -> List[Tuple[int, int]]:
+    """
+    Split the input zarr into many zarrs. Names and sizes can be passed using the split_infos arg.
+
+    Args:
```
oliver-scheel, line 179: Minor, nice to have but not sure if worth it: overload this function to work with only a single argument for size denoting the number / size of each split.

l5kit/l5kit/data/zarr_utils.py
```diff
@@ -1,66 +1,223 @@
...
+    input_dataset.open()
+
+    assert len(split_infos) > 0
+    assert split_infos[-1]["split_size_GB"] == -1, "last split element should have split_size_GB equal to -1"
+
+    # compute the size of the input_dataset in GB
+    size_input_gb = _compute_path_size(input_zarr) / GIGABYTE
+    num_scenes_input = len(input_dataset.scenes)
+
+    # ensure we can fit everything
```
oliver-scheel, line 199: Doesn't the following assert and thus this comment check something more like 'Did the user define splits which are too large?' (kind of the opposite of 'we can fit everything'

lucabergamini: It's probably the same thing seen from two different points of view xD 
I'll change for clarity :)

l5kit/l5kit/data/zarr_utils.py
lucabergamini: mmmmm python doesn't support function overloading easily so I would leave it for another PR maybe

PR #104, opened 2020-08-17
remove build_dataloader

lucabergamini:
Remove `build_dataloader` function from codebase, update notebooks.

Also:
- `open` and `initialize` now return a `ChunkedDataset`
- agents_mask clean is performed when tests start also


l5kit/l5kit/data/zarr_dataset.py
```diff
@@ -114,8 +114,9 @@ def initialize(
 
         self.root.attrs["format_version"] = FORMAT_VERSION
         self.root.attrs["labels"] = PERCEPTION_LABELS
+        return self
```
yye-lyft, line 117: Return `self` is strange

lucabergamini: it's quite common for fluent API interfaces I think. I would leave it as it is, it doesn't have any side effect and avoid us 1 line around all the codebase

PR #106, opened 2020-08-17
Add Metrics

lucabergamini:
- [x] rmse
- [x] rmse tests
- [x] true mode prob
- [x] true mode prob tests (just a couple with random values)
- [x] displacement at t
- [x] displacement at t tests (just a couple with random values)


l5kit/l5kit/evaluation/metrics.py
```diff
@@ -1,23 +1,18 @@
...
     """
-    Compute a negative log-likelihood for the multi-modal scenario.
-    log-sum-exp trick is used here to avoid underflow and overflow, For more information about it see:
-    https://en.wikipedia.org/wiki/LogSumExp#log-sum-exp_trick_for_log-domain_calculations
-    https://timvieira.github.io/blog/post/2014/02/11/exp-normalize-trick/
-    https://leimao.github.io/blog/LogSumExp/
+    Check the shapes of args required by metrics
 
     Args:
         gt (np.ndarray): array of shape (time)x(2D coords)
```
oliver-scheel, line 9: Maybe replacing 'time' with 'future timesteps' or something like that is a bit more usual?

l5kit/l5kit/evaluation/metrics.py
```diff
@@ -32,6 +27,28 @@ def neg_multi_log_likelihood(
...
+
+def neg_multi_log_likelihood(
+    gt: np.ndarray, pred: np.ndarray, confidences: np.ndarray, avails: np.ndarray
+) -> np.ndarray:
+    """
+    Compute a negative log-likelihood for the multi-modal scenario.
+    log-sum-exp trick is used here to avoid underflow and overflow, For more information about it see:
+    https://en.wikipedia.org/wiki/LogSumExp#log-sum-exp_trick_for_log-domain_calculations
+    https://timvieira.github.io/blog/post/2014/02/11/exp-normalize-trick/
+    https://leimao.github.io/blog/LogSumExp/
```
oliver-scheel, line 39: Maybe we can add a bit more comments here? Ie, mention first the formula which is our loss, and then show the modified formula for stability?

l5kit/l5kit/evaluation/metrics.py
```diff
@@ -1,46 +1,293 @@
...
-    log-sum-exp trick is used here to avoid underflow and overflow, For more information about it see:
-    https://en.wikipedia.org/wiki/LogSumExp#log-sum-exp_trick_for_log-domain_calculations
-    https://timvieira.github.io/blog/post/2014/02/11/exp-normalize-trick/
-    https://leimao.github.io/blog/LogSumExp/
+    Check the shapes of args required by metrics
 
     Args:
-        gt (np.ndarray): array of shape (time)x(2D coords)
-        pred (np.ndarray): array of shape (modes)x(time)x(2D coords)
+        truth (np.ndarray): array of shape (timesteps)x(2D coords)
```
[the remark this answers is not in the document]
lucabergamini: ground truth is too long. It doesn't really add anything to truth... we can do it if you're adamant on this but I don't really see the point.

oliver-scheel (dismissed):

oliver-scheel (approved):
LGTM

l5kit/l5kit/tests/evaluation/metrics_test.py
```diff
@@ -1,26 +1,45 @@
...
+)
+
+
+def test_assert_shapes() -> None:
+    num_modes, future_len, num_coords = 4, 12, 2
+
+    gt = np.random.randn(future_len, num_coords)
+    pred = np.random.randn(num_modes, future_len, num_coords)
+    avail = np.ones(future_len)
+    conf = np.random.rand(num_modes)
```
laky, line 23: Do we need the rand / randn above? Could we just use zeros? From a quick read it seems like it shouldn't make a difference. Just a bit worried of using randomness in tests as they may end up flaky.

l5kit/l5kit/evaluation/metrics.py
```diff
@@ -1,46 +1,299 @@
...
+    Return the displacement at timesteps T
+
+    Args:
+        ground_truth (np.ndarray): array of shape (timesteps)x(2D coords)
+        pred (np.ndarray): array of shape (modes)x(timesteps)x(2D coords)
+        confidences (np.ndarray): array of shape (modes) with a confidence for each mode in each sample
+        avails (np.ndarray): array of shape (timesteps) with the availability for each gt timesteps
+
+    Returns:
+        np.ndarray: a (timesteps) numpy array
```
laky, line 137: What is the dimension of this array? It might be good to include, similarly to the input arrays.

stefaniespeichert (commented):
LGTM now

l5kit/l5kit/evaluation/metrics.py
```diff
@@ -1,46 +1,299 @@
...
+        error = np.log(confidences) - 0.5 * np.sum(error, axis=-1)  # reduce timesteps
+
+    # use max aggregator on modes for numerical stability
+    max_value = error.max()  # error are negative at this point, so max() gives the minimum one
+
+    error = np.exp(error - max_value) / np.sum(np.exp(error - max_value))
+    return error
+
+
+def time_displace(
```
laky, line 124: Could this be called `displacements_at_time` or `displacements_at_steps` or something like that? This seems a bit of a cryptic name for a public function.

l5kit/l5kit/evaluation/metrics.py
```diff
@@ -1,46 +1,299 @@
...
+    ground_truth = np.expand_dims(ground_truth, 0)  # add modes
+    avails = avails[np.newaxis, :, np.newaxis]  # add modes and cords
+
+    error = np.sum(((ground_truth - pred) * avails) ** 2, axis=-1)  # reduce coords and use availability
+
+    with np.errstate(divide="ignore"):  # when confidence is 0 log goes to -inf, but we're fine with it
+        error = np.log(confidences) - 0.5 * np.sum(error, axis=-1)  # reduce timesteps
+
+    # use max aggregator on modes for numerical stability
+    max_value = error.max()  # error are negative at this point, so max() gives the minimum one
```
laky, line 118: The above seems the same as the computation for NLL. Consider creating a submethod if it makes sense.

l5kit/l5kit/evaluation/metrics.py
```diff
@@ -1,46 +1,299 @@
 import numpy as np
 
 
-def neg_multi_log_likelihood(
-    gt: np.ndarray, pred: np.ndarray, confidences: np.ndarray, avails: np.ndarray
-) -> np.ndarray:
+def _assert_shapes(ground_truth: np.ndarray, pred: np.ndarray, confidences: np.ndarray, avails: np.ndarray) -> None:
```
laky, line 4: Do we enforce some number of modes for the competition? Might be worthwhile checking that in the computation if yes. Otherwise it seems any number of modes is accepted. Is there perhaps a maximum we should enforce at least?

And I assume that we expect the same number of modes for all predictions, i.e., we can't predict 3 for one situation and 5 for another.

laky (commented):
Mostly ok, 2 questions to consider:

1. Should there be some limits on the number of modes people submit? This could break / run out of memory if someone sends in too large number of predictions.

2. Let's test that this doesn't run out of memory when evaluated on the whole validation set.

PR #109, opened 2020-08-18
move default dataset meta path, add warning for cv2 on macOS

lucabergamini:
Move default dataset meta path into `semantic_map`. As a dataset is tied to it's semantic map it makes sense to have it there

Update: it will be moved yet again as we have decided on a different structure for the dataset

Also, add a note for training on macOS using the `py_satellite`. In this case opencv `warpAffine` breaks because of multi-processing. A quick fix is to disable cv2 multi-processing

### DIFF
--- a/examples/agent_motion_prediction/agent_motion_prediction.ipynb
+++ b/examples/agent_motion_prediction/agent_motion_prediction.ipynb
@@ -159,8 +159,8 @@
    "source": [
     "# Training\n",
     "\n",
-    "note: if you're on MacOS and you're using `py_satellite` rasterizer, you may need to disable opencv multiprocessing by adding:\n",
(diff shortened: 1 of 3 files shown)


examples/agent_motion_prediction/agent_motion_prediction.ipynb
```diff
@@ -157,7 +157,10 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
-    "# Training"
+    "# Training\n",
+    "\n",
+    "note: if you're on MacOS and you're using `py_satellite` rasterizer, you may need to disable opencv multiprocessing by adding:\n",
```
oliver-scheel, line 162: Minor: i would remove the second 'you're, 
+ cause -> caused

oliver-scheel (approved):
Minors, LGTM

oliver-scheel (approved):

PR #110, opened 2020-08-19
clarify dataset structure in readme

lucabergamini:


lucabergamini:
close as we data layout will change in v1.1. Move some of these changes in #112

PR #111, opened 2020-08-19
relax version constraint for pytorch, to not conflict with Jadoo

epraun:
See also, 
https://github.com/lyft/avresearch/pull/331
https://github.com/lyft/avresearch/pull/316


l5kit/setup.py
```diff
@@ -20,8 +20,8 @@
         "pymap3d",
         "scipy",
         "setuptools",
-        "torch==1.5.1",
-        "torchvision==0.6.1",
+        "torch>=1.5.0",
```
lucabergamini, line 23: this will install torch 1.6.0 which we haven't tested yet. Can we set it to >=1.5.0,<1.6.0?

epraun: thanks! done

lucabergamini (approved):
LGTM

PR #115, opened 2020-08-21
add zarr insights

lucabergamini:
Add some insights on how to work with zarr through our `ChunkedDataset`

### DIFF
--- a/data_format.md
+++ b/data_format.md
@@ -94,7 +94,7 @@ Some other zarr benefits are:
 * The metadata (e.g. dtype, chunk size, compression type) is stored inside the zarr dataset too. If one day you decide to change your chunk size, you can still read the older datasets without changing any code.
 
 ## 2020 Lyft Competition Dataset format
-The 2020 Lyft competition dataset is stored in three structured arrays: `scenes`, `frames` and `agents`.
+The 2020 Lyft competition dataset is stored in four structured arrays: `scenes`, `frames`, `agents` and `tl_faces`.
 
 ### Scenes
 A scene is identified by the host (i.e. which car was used to collect it) and a start and end time. It consists of multiple frames (= discretized measurements), in the scene datatype we store the start and end index in the `frames` array described below that correspond to this scene.
@@ -109,12 +109,13 @@ SCENE_DTYPE = [
 ```
 
 ### Frames
-A frame consists of information about the ego vehicle (e.g. where it was at that time), a timestamp, and a reference to the agents in that frame. Because there may be multiple agents observed we store the start and the end index.
+A frame consists of information about the ego vehicle (e.g. where it was at that time), a timestamp, and a reference to the agents and traffic light faces in that frame. Because there may be multiple agents/faces observed we store the start and the end index.
 
 ```python
 FRAME_DTYPE = [
     ("timestamp", np.int64),
     ("agent_index_interval", np.int64, (2,)),
+    ("traffic_light_faces_index_interval", np.int64, (2,)),
     ("ego_translation", np.float64, (3,)),
     ("ego_rotation", np.float64, (3, 3)),
 ]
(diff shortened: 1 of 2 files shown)


oliver-scheel:
Not part of the PR, but Im going the whole doc now: ll 28, print(my_arr["color"]) - if the following comment is supposed to be the output of this command, this is wrong - label is not printed.

data_format.md
```diff
@@ -94,10 +94,16 @@ Some other zarr benefits are:
 * The metadata (e.g. dtype, chunk size, compression type) is stored inside the zarr dataset too. If one day you decide to change your chunk size, you can still read the older datasets without changing any code.
 
 ## 2020 Lyft Competition Dataset format
-The 2020 Lyft competition dataset is stored in three structured arrays: `scenes`, `frames` and `agents`.
+The 2020 Lyft competition dataset is stored in four structured arrays: `scenes`, `frames`, `agents` and `tl_faces`.
+
+Note: in the following all `_interval` fields assume that information is stored consecutively in the arrays.
+This means that if `frame_index_interval` for `scene_0` are `(0, 100)`, frames from `scene_1` will start from index 100 in the frames array.
```
oliver-scheel, line 100: is (0, 100)

data_format.md
```diff
@@ -94,10 +94,16 @@ Some other zarr benefits are:
...
 
 ### Scenes
-A scene is identified by the host (i.e. which car was used to collect it) and a start and end time. It consists of multiple frames (= discretized measurements), in the scene datatype we store the start and end index in the `frames` array described below that correspond to this scene.
+A scene is identified by the host (i.e. which car was used to collect it) and a start and end time. 
+It consists of multiple frames (=snapshots at discretized time intervals). 
+The scene datatype stores references to its corresponding frames in terms of the start and end index within the frames array (described below). 
+The frames in between these indices all correspond to the scene (Including start index, excluding end index).
```
oliver-scheel, line 106: including

data_format.md
```diff
@@ -109,19 +115,31 @@ SCENE_DTYPE = [
 ```
 
 ### Frames
-A frame consists of information about the ego vehicle (e.g. where it was at that time), a timestamp, and a reference to the agents in that frame. Because there may be multiple agents observed we store the start and the end index.
+A frame captures all information that was observed at a time. This includes
+
+- the timestamp, which the frame describes;
+- data about the ego vehicle itself such as rotation and position;
+- a reference to the other agents (vehicles, cyclists and pedestrians) that were captured by the ego's sensors;
+- a reference to all traffic light faces for all visible lanes.
```
oliver-scheel, line 123: I actually do not think traffic light faces is a common English word, so one could wonder what that is upon arriving here - it is solved in the paragraph below, so not sure how important. One could just say 'traffic lights' here or add 'see below' to help the reader here?

data_format.md
```diff
@@ -109,19 +115,31 @@ SCENE_DTYPE = [
...
+
+The properties for both agents and traffic light faces are stored in their two respective arrays. 
+The frame contains only pointers to these stored objects in terms of a start and an end index to these arrays (again, start in included while end excluded).
```
oliver-scheel, line 126: Here's somehow twice pointer TO - maybe change to 'The frame contains only pointers to these stored objects in terms of a start and an end index in these arrays' or just remove 'to these arrays.'

oliver-scheel: Start IS included?

data_format.md
```diff
@@ -133,3 +151,60 @@ AGENT_DTYPE = [
     ("label_probabilities", np.float32, (len(LABELS),)),
 ]
 ```
+
+### Traffic Light Faces
+Note: we refer to traffic light bulbs (e.g. the red light bulb of a specific traffic light) as `faces` in L5Kit.
+For the full list of available types for a bulb please consult our [protobuf map definition](https://github.com/lyft/l5kit/blob/20ab033c01610d711c3d36e1963ecec86e8b85b6/l5kit/l5kit/data/proto/road_network.proto#L615).
+
+Our semantic map holds static information about the world only. This means it has a list of all traffic lights, but not information about how their status changes over time.
```
oliver-scheel, line 159: but NO information

data_format.md
```diff
@@ -133,3 +151,60 @@ AGENT_DTYPE = [
...
+This dynamic information is instead stored in this array.
+Each array's element has a unique id to link it to the semantic map, a status (if `>0` then the face is active) and a reference to its parent traffic light.
```
oliver-scheel, line 161: Could think of giving a concrete example here, ie >0 means red bulb is one, =0 for green bulb

data_format.md
```diff
@@ -133,3 +151,60 @@ AGENT_DTYPE = [
...
+    ("face_id", "<U16"),
+    ("traffic_light_id", "<U16"),
+    ("traffic_light_face_status", np.float32, (len(TL_FACE_LABELS,))),
+]
+```
+
+## Working with the zarr format
+
+### The ChunkedDataset
+The `ChunkedDataset` (`l5kit.data.zarr_dataset`) is the first interface between raw data on the disk and python accessible information.
```
oliver-scheel, line 174: Python

data_format.md
```diff
@@ -133,3 +151,60 @@ AGENT_DTYPE = [
...
+This layer is very thin, and it provides the four arrays mapped from the disk. When one of these array is indexed (or sliced):
+- `zarr` identifies the chunk(s) to be loaded;
+- the chunk is decompressed on the fly;
+- a numpy array copy is returned; 
+
+The `ChunkedDataset` also provides a LRUcache; but it works on [compressed chunks only](https://github.com/zarr-developers/zarr-python/issues/278).
```
oliver-scheel, line 180: an LRUcache

data_format.md
```diff
@@ -133,3 +151,60 @@ AGENT_DTYPE = [
...
+dt = ChunkedDataset("<path>").open()
+centroids= []
+for idx in range(10_000):
+    centroid = dt.agents[idx]["centroid"]
+    centroids.append(centroid)
+```
+
+However, in this implementation **we are decompressing the same chunk (or two) 10_000 times!**
+
+If we rewrite it as:
```
oliver-scheel, line 200: id remove the colon here and then continue with lower case we

data_format.md
stefaniespeichert: second this! It was a bit confusing to me at first as well!

stefaniespeichert (commented):
Oliver left some good comments here so once addressed we're good to go!

oliver-scheel:
> Oliver left some good comments here so once addressed we're good to go!

adressed them, if you could take another look @stefaniespeichert and approve if ok? :)

stefaniespeichert (approved):
lgtm now :)

PR #120, opened 2020-08-25
define metric signature and ensure mypy check

lucabergamini:
Define a metric signature using Typing, so that mypy can warn us if we use a metric with a different signature


stefaniespeichert (approved):
lgtm

PR #121, opened 2020-08-25
add support for warnings

lucabergamini:
replace prints with python warnings of different types

### DIFF
--- a/l5kit/l5kit/geometry/transform.py
+++ b/l5kit/l5kit/geometry/transform.py
@@ -1,4 +1,3 @@
-import warnings
 from typing import Optional, Sequence, Tuple, Union, cast
 
 import numpy as np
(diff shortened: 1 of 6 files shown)


l5kit/l5kit/geometry/transform.py
```diff
@@ -111,6 +112,8 @@ def transform_points(points: np.ndarray, transf_matrix: np.ndarray) -> np.ndarra
     Returns:
         np.ndarray: array of shape (N,2) for 2D input points, or (N,3) points for 3D input points
     """
+    if len(points.shape) == 1:
```
oliver-scheel, line 115: Why was this changed?

lucabergamini: because we're deprecating the function `transform_point` and this one will replace it. I should probably ensure the final shape is preserved though, what do you think?

lucabergamini: I've removed this from the PR. I'll start a dedicated one targeting this one

oliver-scheel (approved):
LGTM

PR #124, opened 2020-08-26
Add selective path for spawn method

lucabergamini:
Since this affects macos only, I've restricted it to it. Also, the method is not available on Windows so it was causing issues there


oliver-scheel (approved):
LGTM

PR #126, opened 2020-08-26
Lberg/pipenv

lucabergamini:
Add pipenv as dependencies manager


.github/workflows/docs.yml
```diff
@@ -20,13 +20,15 @@ jobs:
     - name: Install dependencies
       working-directory: l5kit
       run: |
-        python -m pip install --upgrade pip
-        pip install -r requirements.txt
+        pip install --upgrade pip
+        pip install pipenv
+        pipenv install --dev
+        pip install .
```
giacomodabisias, line 26: what is this last line for?

.github/workflows/docs.yml
```diff
@@ -20,13 +20,15 @@ jobs:
     - name: Install dependencies
       working-directory: l5kit
       run: |
-        python -m pip install --upgrade pip
-        pip install -r requirements.txt
+        pip install --upgrade pip
+        pip install pipenv
+        pipenv install --dev
```
giacomodabisias, line 25: Always use `pipenv sync` to make sure that you install packages from the lock file.
No need for `--dev` since you are not developing here

l5kit/.mypy.ini
```diff
@@ -6,10 +6,4 @@ disallow_untyped_defs = True
...
 
 [mypy-l5kit/data/proto/*]
-ignore_errors = True
-
-[metadata]
-description-file = README.md
-
-[tool:pytest]
-xfail_strict=true
+ignore_errors = True
```
giacomodabisias, line 9: missing new line

l5kit/Pipfile
```diff
@@ -0,0 +1,42 @@
...
+protobuf = ">=3.12.2"
+torch = ">=1.5.0,<1.6.0"
+torchvision = ">=0.6.0,<0.7.0"
+ipython = "<=7.16.1"
+
+[requires]
+python_version = "3.8"
+
+[pipenv]
+allow_prereleases = true
```
giacomodabisias, line 42: no need for this since you pinned the black version and nothing else looks like a pre-release

l5kit/Pipfile
```diff
@@ -0,0 +1,42 @@
...
+pyyaml = "*"
+matplotlib = "*"
+imageio = "*"
+notebook = "*"
+ptable = "*"
+ipywidgets = "*"
+protobuf = ">=3.12.2"
+torch = ">=1.5.0,<1.6.0"
+torchvision = ">=0.6.0,<0.7.0"
+ipython = "<=7.16.1"
```
giacomodabisias, line 36: This seems dangerous. Was there any incompatible change lately? If yes please add a comment

giacomodabisias:
Small NIT: Always add a description to a PR to give a high overview of what is going on :)

.github/workflows/docs.yml
lucabergamini: we read the version from the `__init__` in l5kit, so we need it to be installed here..

.github/workflows/docs.yml
lucabergamini: Currently our doc tools are installed as `dev` dependencies so they need `--dev`, but maybe that's not a best practice?

l5kit/Pipfile
lucabergamini: It's autogenerated, should I still remove it?

l5kit/Pipfile
lucabergamini: No, but macOS has already the 7.17 while linux doesn't. If I generated the lock from MacOS without capping the version CI workers won't be able to install from the lock..

.github/workflows/docs.yml
giacomodabisias: not sure I understand. What is this installing?

l5kit/Pipfile
giacomodabisias: You mean that the latest versions here https://pypi.org/project/ipython/#history are only for Mac? That seems very strange. I would assume that thing are always first available for linux and later for mac, if  this actually happens.

l5kit/Pipfile
giacomodabisias: How are you autogenerating the Pipfile?

.github/workflows/docs.yml
lucabergamini: l5kit itself from the `setup.py`

l5kit/Pipfile
lucabergamini: No sorry I mean those lines were autogenerated after installing black.

l5kit/Pipfile
lucabergamini: I've checked now and it seems like that issue has been fixed. Let me try to remove the line and see if it passes on CI machines

.github/workflows/docs.yml
giacomodabisias: but if you add l5kit to the pipfile there is no need for this, right?

.github/workflows/docs.yml
giacomodabisias: docs are not part of dev :)
Dev packages are packages needed for development and documentation is not part of that

l5kit/Pipfile
giacomodabisias: no need for that since you specify manually the black version as a pre release

l5kit/Pipfile
giacomodabisias: nice :)

.github/workflows/docs.yml
lucabergamini: I agree, but they are also not part of the library itself right? Should they be install together with it regardless?

.github/workflows/docs.yml
lucabergamini: ok that makes sense, should I add it in editable mode then?

l5kit/Pipfile
lucabergamini: checked again, newer ipython are [only 3.7+](https://ipython.readthedocs.io/en/stable/whatsnew/version7.html) . I can generate the lock from a 3.6 python to ensure the locked version is 3.6-compatible, is that fine with you? The risk there is obviously someone updating the lock in the future from 3.8  or 3.7.

Alternatively, we could drop support for 3.6

.github/workflows/docs.yml
giacomodabisias: yep, I would add it there so that pipenv manages everything. There should't be the need to run any pip command

.github/workflows/docs.yml
giacomodabisias: the dev packages should be installed when you want to do development :) In CI there is no development so there should be no --dev. Whatever is needed to produce some artifacts or to test something should not be dev

l5kit/Pipfile
giacomodabisias: why is ipython in the pipfile? 
is there a dependency on it?

lucabergamini: jupyter has a dependency on ipython. If I let pipenv to manage it it will have a different behaviour depending on the python version you use to generate the lock:
- python 3.6 -> lock 7.16
- python 3.7, 3.8 -> lock 7.18 or the most recent one

The issue here is that ipython has dropped support for 3.6 between two versions, what is the best practise here?

Jimexist:
I wonder if we can use [Poetry](https://python-poetry.org) instead? Since there's already pyproject.toml file in place.

lucabergamini:
> I wonder if we can use [Poetry](https://python-poetry.org) instead? Since there's already pyproject.toml file in place.

Big fan of Poetry here (there is even a branch which supports it already), but for internal reason we have to use pipenv...

giacomodabisias:
@Jimexist  Poetry could be the future of dependency management, but currently it is still at version 1.0.10. 
i'd wait a bit longer before using something which is so fresh and use something more reliable and stable in the meantime (not saying that pipenv is perfect, but it has been around for longer)

l5kit/Pipfile
giacomodabisias: jupyter should resolve its dependencies on its own. II don't think you need to add the internal dependencies.
In general I don't think its a great idea to support several python versions. What is the advantage? You can just stick to one version.

lucabergamini:
Also, Poetry doesn't support `setup.py` which is still the only supported way of installing a package from pip (this will likely change in the future)

l5kit/Pipfile
lucabergamini: > jupyter should resolve its dependencies on its own. II don't think you need to add the internal dependencies.

It's a locking issue here. If we lock from 3.8, ipython will be locked at 7.18 which is not available on 3.6 in CI.



> In general I don't think it's a great idea to support several python versions. What is the advantage? You can just stick to one version.

Well I guess one advantage is that if you're on ubuntu 18.04LTS (default python is still 3.6.8 there) you don't have to install a new interpreter just to use our package. Apart from that, I don't see any particular issues in dropping support for 3.6 and 3.7. It will also make maintenance easier. So should we just aim to support 3.8 in the end @giacomodabisias ?

Jimexist:
> Also, Poetry doesn't support `setup.py` which is still the only supported way of installing a package from pip (this will likely change in the future)

agree that in this case Pipenv is a better choice.

l5kit/setup.py
```diff
@@ -31,9 +31,9 @@
         "ipywidgets"
     ],
     extras_require={
-        "dev": ["pytest==5.4.3", "mypy==0.782", "setuptools", "twine", "wheel", "pytest-cov==2.10.0", "flake8==3.8.3",
-                "black==19.10b0", "isort==5.0.4", "Sphinx==3.1.1", "sphinx-rtd-theme==0.5.0", "recommonmark==0.6.0",
-                "pre-commit==2.5.1"]
+        "dev": ["pytest>=5.4.3", "mypy>=0.782", "setuptools", "twine", "wheel", "pytest-cov>=2.10.0", "flake8>=3.8.3",
```
giacomodabisias, line 34: why do you need `mypy>=0.782` and "pytest-cov>=2.10.0" and "flake8>=3.8.3" and "isort>=5.0.4" ?
These are dev packages and you should be able to always use the latest version right? 
Is there any specific thing happening with the pinned versions?

lucabergamini: There are issues happening with previous versions (so for example mypy 0.77). I guess this is not really necessary though, as the behaviour should be the same as `*`. I'll remove it

Jimexist:
i am all for dropping py3.6

lucabergamini:
> i am all for dropping py3.6

Already included in this PR :) We'll stop testing it in CI :)

.github/workflows/l5kit-lock.yml
```diff
@@ -0,0 +1,42 @@
...
+    - name: Set up Python ${{ matrix.python-version }}
+      uses: actions/setup-python@v1
+      with:
+        python-version: ${{ matrix.python-version }}
+    - name: Install dependencies
+      working-directory: l5kit
+      run: |
+        pip install --upgrade pip
+        pip install pipenv
+        pipenv sync -d
```
giacomodabisias, line 26: is there no need for a `wait` here since all steps run in parallel?

.github/workflows/l5kit-lock.yml
```diff
@@ -0,0 +1,42 @@
...
+
+jobs:
+  CI-build:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        python-version: [3.7, 3.8]
+
+    steps:
+    - uses: actions/checkout@v2
```
giacomodabisias, line 16: can we indent all steps for readability?

README.md
```diff
@@ -115,10 +115,25 @@ git clone https://github.com/lyft/l5kit.git ./
...
 ### 3. Install L5Kit
+
+#### 3.1 Deterministic Build (Suggested)
+We support deterministic build through [pipenv](https://pipenv-fork.readthedocs.io/en/latest/).
+
+Once you've installed pipenv (or made it available in your env) run: 
 ```shell
 cd l5kit
-pip install -r requirements.txt
+pipenv sync -d
```
giacomodabisias, line 125:
```suggestion
pipenv sync --dev
```
for readability

.github/workflows/l5kit-lock.yml
giacomodabisias, line 26:
```suggestion
        pipenv sync --dev
```
for reaedability

README.md
```diff
@@ -115,10 +115,25 @@ git clone https://github.com/lyft/l5kit.git ./
...
 ```
+This will install all dependencies (`-d` includes dev-packages too) from the lock file.
```
giacomodabisias, line 127:
```suggestion
This will install all dependencies (`--ddev` includes dev-packages too) from the lock file.
```

l5kit/Pipfile
```diff
@@ -0,0 +1,36 @@
...
+pyyaml = "*"
+matplotlib = "*"
+imageio = "*"
+notebook = "*"
+ptable = "*"
+ipywidgets = "*"
+protobuf = ">=3.12.2"
+torch = ">=1.5.0,<1.6.0"
+torchvision = ">=0.6.0,<0.7.0"
+l5kit = {editable = true, path = "."}
```
giacomodabisias, line 36: I'd move this to thee top of the list

giacomodabisias, line 36:
```suggestion
l5kit = {editable = true, path = "./l5kit"}
```

giacomodabisias (commented):
Nice progress!
Almost there, some more minor comments

.github/workflows/l5kit-lock.yml
lucabergamini: As far as I know `jobs` are run in parallel in a workflow, but `steps` and `run` are sequential. I would be very surprised if that was not the case ad we've never experienced any sync issues.
https://resources.github.com/whitepapers/GitHub-Actions-Cheat-sheet/ also seems to confirm it (I haven't found a more official doc that explains it though)

giacomodabisias: Ah wait, I aws looking at buildkite xD never mind!

l5kit/Pipfile
lucabergamini: I'm not able to `pipenv lock` anymore if I add this:
`pipenv.vendor.requirementslib.exceptions.RequirementError: Failed parsing requirement from './l5kit'` 
 
Why should it point to `./l5kit` and not `.` ?

giacomodabisias (dismissed):
Great Job :)

l5kit/Pipfile
lucabergamini: I think [it should point](https://stackoverflow.com/questions/53378416/what-does-pipenv-install-e-do-and-how-to-use-it) to where the `setup.py` is (otherwise pip doesn't know how to build it). If that's the case it makes sense to leave it as `.` here, what do you think @giacomodabisias

giacomodabisias: Ah your pipfile is in the L5kit folder, I see. Usually you have the pipfile and pipfile.lock one folder up, since you might create new packages in the same repo and you can use the pipfile to manage all of them if they work together.
I suggest moving the files up one folder to the main folder of the repo

lucabergamini: mmm...given that this is not the case for this repo (not planned anyway) I would leave it there. The reason is also that all other configuration files (and `setup.py` itself) are already placed there. Moving only this one one level up would create some confusion about where to run what imo (e.g. install and add packages vs running tests, lint or installing from `setup.py`).

giacomodabisias: up to you :) I am fine also with the current structure.
Notice though that there would be no confusion since tests, linting etc..are all independent from the file location.
Setup.py should be in the package as it is now and should not be moved :)

lucabergamini: > Notice though that there would be no confusion since tests, linting etc..are all independent from the file location.

They still need to read their config files though, otherwise they run with default configs.

I'll leave it as it is for now but I'll consider it for the future :) Can I have your final blessing @giacomodabisias 😄 ?

giacomodabisias: sure, go for it

yye-lyft (approved):

Jimexist:
thanks for the change. do we know if we want to cut a new release after this?

lucabergamini:
> thanks for the change. do we know if we want to cut a new release after this?

Likely after #150 and #151 are merge, hopefully by middle of the next week we should have a new release tested :)

PR #128, opened 2020-08-30
Correct the link to the possible defined labels

witignite:
I think the link to the *possible defined labels* is probably wrong, so I change it so that it points to: https://github.com/lyft/l5kit/blob/master/l5kit/l5kit/data/labels.py


lucabergamini:
Hi @witignite 


> is probably wrong

Definitely wrong, thanks for picking this up :)

lucabergamini (approved):
LGTM

PR #133, opened 2020-09-01
Fix box_rasterizer output shape

lucabergamini:
`box_rasterizer` width and height were swapped in multiple places, making it failing for rectangular shapes.

This PR fixes it but it seems like `py_satellite` has unexpected behaviours with rectangular rasters:

`py_satellite` with (128, 224):
Screenshot 2020-09-01 at 19 02 16
`py_satellite` with (224, 128):
Screenshot 2020-09-01 at 19 03 57

### DIFF
--- a/l5kit/l5kit/rasterization/box_rasterizer.py
+++ b/l5kit/l5kit/rasterization/box_rasterizer.py
@@ -49,9 +49,9 @@ def draw_boxes(
         np.ndarray: the image with agents rendered. RGB if color RGB, otherwise GRAY
     """
     if isinstance(color, int):
-        im = np.zeros(raster_size, dtype=np.uint8)
+        im = np.zeros((raster_size[1], raster_size[0]), dtype=np.uint8)
     else:
-        im = np.zeros(raster_size + (3,), dtype=np.uint8)
+        im = np.zeros((raster_size[1], raster_size[0], 3), dtype=np.uint8)
 
     box_world_coords = np.zeros((len(agents), 4, 2))
     corners_base_coords = np.asarray([[-1, -1], [-1, 1], [1, 1], [1, -1]])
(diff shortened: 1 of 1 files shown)


oliver-scheel (commented):
LGTM - what about the satellite rasterizer though?

lucabergamini:
> LGTM - what about the satellite rasterizer though?

I'll start another PR to fix that (once I understand what's wrong with it xD)

yye-lyft:
> LGTM - what about the satellite rasterizer though?

You forgot to approve the PR...

l5kit/l5kit/rasterization/box_rasterizer.py
```diff
@@ -49,9 +49,9 @@ def draw_boxes(
         np.ndarray: the image with agents rendered. RGB if color RGB, otherwise GRAY
     """
     if isinstance(color, int):
-        im = np.zeros(raster_size, dtype=np.uint8)
+        im = np.zeros((raster_size[1], raster_size[0]), dtype=np.uint8)
```
yye-lyft, line 52: can we avoid swapping box_rasterizer width and height? This is counter-intuitive

lucabergamini: Our codebase is XY, but numpy wants shape as YX so I don't really know how we can avoid that, any idea?

yye-lyft: okay I see. Yeah, I do not have good suggestion

yye-lyft (approved):

Yannik-H:
Is rectangular shape available now? I tried to use raster size=(400,200), and an exception occurred saying that the shape should be the same. By the way, the l5kit is 1.0.6. Is the version of l5kit too old to build a rasterizer for rectangular shape?

lucabergamini:
>  By the way, the l5kit is 1.0.6. Is the version of l5kit too old to build a rasterizer for rectangular shape?

Yup, you should be on 1.1.0 onwards for rectangular shape support :)

PR #134, opened 2020-09-02
speed up get_frame_indices by avoiding zarr access

lucabergamini:
Speed up `get_frame_indices` by replacing `zarr` access with indexing over `cumulative_sizes_agents`. This should give us a noticeable performance boost.

### DIFF
--- a/l5kit/l5kit/dataset/agent.py
+++ b/l5kit/l5kit/dataset/agent.py
@@ -152,11 +152,12 @@ class AgentDataset(EgoDataset):
         Returns:
             np.ndarray: indices that can be used for indexing with __getitem__
         """
-        frames = self.dataset.frames
-        assert frame_idx < len(frames), f"frame_idx {frame_idx} is over len {len(frames)}"
+        assert frame_idx < len(self.dataset.frames), f"frame_idx {frame_idx} is over len {len(self.dataset.frames)}"
 
-        agent_slice = get_agents_slice_from_frames(frames[frame_idx])
+        # avoid accessing zarr here as we already have the information in `cumulative_sizes_agents`
+        agent_end = self.cumulative_sizes_agents[frame_idx]
+        agent_start = self.cumulative_sizes_agents[frame_idx - 1] if frame_idx > 0 else 0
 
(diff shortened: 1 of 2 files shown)


l5kit/l5kit/dataset/agent.py
```diff
@@ -152,11 +152,12 @@ def get_frame_indices(self, frame_idx: int) -> np.ndarray:
...
             np.ndarray: indices that can be used for indexing with __getitem__
         """
-        frames = self.dataset.frames
-        assert frame_idx < len(frames), f"frame_idx {frame_idx} is over len {len(frames)}"
+        assert frame_idx < len(self.dataset.frames), f"frame_idx {frame_idx} is over len {len(self.dataset.frames)}"
 
-        agent_slice = get_agents_slice_from_frames(frames[frame_idx])
+        # avoid accessing zarr here as we already have the information in `cumulative_sizes_agents`
+        agent_end = self.cumulative_sizes_agents[frame_idx]
+        agent_start = self.cumulative_sizes_agents[frame_idx - 1] if frame_idx > 0 else 0
```
laky, line 159: Very nit, I'd swap the lines as start is usually before end :)

laky (dismissed):
LGTM, thanks a lot! Is there a test for the function? If not, it might be good to add one to make sure we don't break things when we merge this!

lucabergamini:
> LGTM, thanks a lot! Is there a test for the function? If not, it might be good to add one to make sure we don't break things when we merge this!

I think we currently have no tests for this! let me add some then!

laky (approved):
Thanks for the tests! Might be nice to name them s.t. they describe the case they're testing a bit more. It is a bit hard to read and understand what is exactly being tested. Whether it is some specific case and what it is.

lucabergamini:
> Might be nice to name them s.t. they describe the case they're testing a bit more

I think we should start considering adding docstring also for tests maybe. The names here are quite clear to me (test+function_name+dataset_type). The issue is that without context is quite hard to understand what's actually going on in almost all our tests.

I'll merge this now but we should discuss whether to add docstrings also for this module

PR #135, opened 2020-09-02
Ensure Dataset has TLs, add option to disable them

lucabergamini:
This PR includes two components:

- Ensure we're loading a dataset with TLs;
- Allow to disable TLs from config

### DIFF
--- a/examples/agent_motion_prediction/agent_motion_config.yaml
+++ b/examples/agent_motion_prediction/agent_motion_config.yaml
@@ -40,6 +40,9 @@ raster_params:
   # one of the classes we care about (cars, bikes, peds, etc.), >=1.0 filter all other agents.
   filter_agents_threshold: 0.5
 
+  # if to disable completely traffic light faces in the semantic rasterizer
+  disable_traffic_light_faces: False
+
 ###################
 ## Data loader options
 train_data_loader:
--- a/examples/agent_motion_prediction/agent_motion_config.yaml
+++ b/examples/agent_motion_prediction/agent_motion_config.yaml
@@ -40,7 +40,7 @@ raster_params:
   # one of the classes we care about (cars, bikes, peds, etc.), >=1.0 filter all other agents.
   filter_agents_threshold: 0.5
 
-  # whether to disable completely traffic light faces in the semantic rasterizer
+  # whether to completely disable traffic light faces in the semantic rasterizer
   disable_traffic_light_faces: False
 
 ###################
--- a/examples/agent_motion_prediction/agent_motion_config.yaml
+++ b/examples/agent_motion_prediction/agent_motion_config.yaml
@@ -40,7 +40,7 @@ raster_params:
   # one of the classes we care about (cars, bikes, peds, etc.), >=1.0 filter all other agents.
   filter_agents_threshold: 0.5
 
-  # if to disable completely traffic light faces in the semantic rasterizer
+  # whether to disable completely traffic light faces in the semantic rasterizer
   disable_traffic_light_faces: False
 
 ###################
--- a/examples/visualisation/visualisation_config.yaml
+++ b/examples/visualisation/visualisation_config.yaml
@@ -40,6 +40,9 @@ raster_params:
   # one of the classes we care about (cars, bikes, peds, etc.), >=1.0 filter all other agents.
   filter_agents_threshold: 0.5
 
+  # if to disable completely traffic light faces in the semantic rasterizer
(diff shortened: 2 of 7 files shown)


examples/agent_motion_prediction/agent_motion_config.yaml
```diff
@@ -40,6 +40,9 @@ raster_params:
   # one of the classes we care about (cars, bikes, peds, etc.), >=1.0 filter all other agents.
   filter_agents_threshold: 0.5
 
+  # if to disable completely traffic light faces in the semantic rasterizer
```
oliver-scheel, line 43: whether

examples/visualisation/visualisation_config.yaml
```diff
@@ -40,6 +40,9 @@ raster_params:
   # one of the classes we care about (cars, bikes, peds, etc.), >=1.0 filter all other agents.
   filter_agents_threshold: 0.5
 
+  # if to disable completely traffic light faces in the semantic rasterizer
```
oliver-scheel, line 43: see above

l5kit/l5kit/data/zarr_dataset.py
```diff
@@ -137,11 +137,7 @@ def open(self, mode: str = "r", cached: bool = True, cache_size_bytes: int = int
         self.frames = self.root[FRAME_ARRAY_KEY]
         self.agents = self.root[AGENT_ARRAY_KEY]
         self.scenes = self.root[SCENE_ARRAY_KEY]
-        try:
-            self.tl_faces = self.root[TL_FACE_ARRAY_KEY]
-        except KeyError:
-            print(f"{TL_FACE_ARRAY_KEY} not found in {self.path}! Traffic lights will be disabled")
-            self.tl_faces = np.empty((0,), dtype=TL_FACE_DTYPE)
+        self.tl_faces = self.root[TL_FACE_ARRAY_KEY]
```
oliver-scheel, line 140: So we just crash? Wasn't the previous version better?

oliver-scheel (dismissed):
Minors, otherwise LGTM

l5kit/l5kit/data/zarr_dataset.py
lucabergamini: I agree; It was better for retro-compatibility, but the thing is that we want to completely move to traffic light datasets only. So it makes sense to hard crash here imo

examples/agent_motion_prediction/agent_motion_config.yaml
```diff
@@ -40,6 +40,9 @@ raster_params:
   # one of the classes we care about (cars, bikes, peds, etc.), >=1.0 filter all other agents.
   filter_agents_threshold: 0.5
 
+  # whether to disable completely traffic light faces in the semantic rasterizer
```
oliver-scheel, line 43: Minor: to be exact still had to be 'whether to completely disable traffic light faces in the semantic rasterizer' or 'whether to disable traffic light faces in the semantic rasterizer completely'

examples/visualisation/visualisation_config.yaml
```diff
@@ -40,6 +40,9 @@ raster_params:
   # one of the classes we care about (cars, bikes, peds, etc.), >=1.0 filter all other agents.
   filter_agents_threshold: 0.5
 
+  # whether to disable completely traffic light faces in the semantic rasterizer
```
oliver-scheel, line 43: see above

l5kit/l5kit/tests/artefacts/config.yaml
```diff
@@ -36,3 +36,6 @@ raster_params:
   # e.g. 0.0 include every obstacle, 0.5 show those obstacles with >0.5 probability of being
   # one of the classes we care about (cars, bikes, peds, etc.), >=1.0 filter all other agents.
   filter_agents_threshold: 0.5
+
+  # whether to disable completely traffic light faces in the semantic rasterizer
```
oliver-scheel, line 40: see above

oliver-scheel (dismissed):
LGTM

l5kit/l5kit/data/zarr_dataset.py
lucabergamini: I have restored this with a stronger warning. For now we will still support loading a zarr without traffic lights but not in the rest of the workflow (e.g. you won't be able to use it in an EgoDataset). In this way we can still inspect raw legacy data if we want to

oliver-scheel (approved):

PR #139, opened 2020-09-08
lessen the max version requirement for torch and torchvision

Jimexist:
now that we have a newer version.

this aligns with the version used in Colab so that people don't need to revert back to a lower version.

for most of the APIs used in this repo, there are no back incompatible changes anyway.


lucabergamini:
Hi @Jimexist, I see two issues here:
- We still need to test 1.6 properly. It's passing tests so that's a good start already :);
- Even if we update this, It's still not 100% clear what the workflow for updating in Kaggle will be. Are people already using a master version of l5kit or the provided release as far as you know?

Jimexist:
> Hi @Jimexist, I see two issues here:
> 
> * We still need to test 1.6 properly. It's passing tests so that's a good start already :);
> * Even if we update this, It's still not 100% clear what the workflow for updating in Kaggle will be. Are people already using a master version of l5kit or the provided release as far as you know?

I guess the only way to effectively test this is to either add a pre-release tag so people can try out or put some examples code within CI so it's automatically picked up and tested end to end.

Jimexist:
@lucabergamini if you trust me ;-), I have tried to run the example baseline model from within Colab and saved a copy of gist here https://gist.github.com/Jimexist/d67a36236e604acc923d6040f92f6233

Basically since l5kit depends on torch 1.5 it will try to uninstall the 1.6 used in Colab and re-install torch 1.5 but sadly this will render the GPU or TPU unusable.

If I instead tried to install `l5kit` with `--no-deps` but then manually install all its deps (except for torch 1.5) it'll work.

lucabergamini:
> uninstall the 1.6 used in Colab

Just to clarify: is 1.5 an issue in kaggle kernels or colab (TPU) only?

> re-install torch 1.5 but sadly this will render the GPU or TPU unusable

Why is that? I'm just curious about it :) 

I guess the best solution would be to:
- rest it a little bit on our side;
- relax the requirement and ship a new release;

Jimexist:
> > uninstall the 1.6 used in Colab
> 
> Just to clarify: is 1.5 an issue in kaggle kernels or colab (TPU) only?
> 
> > re-install torch 1.5 but sadly this will render the GPU or TPU unusable
> 
> Why is that? I'm just curious about it :)
> 
> I guess the best solution would be to:
> 
> * rest it a little bit on our side;
> * relax the requirement and ship a new release;

no I guess there's no issue with 1.5, just that when you try to `pip install l5kit` in Colab (there's no Kaggle here), it'll install a CPU only version.

Jimexist:
> relax the requirement and ship a new release;

this actually makes sense. I guess you can drop the `< 1.6` and `< 0.7` for `torch` and `torchvision` without forcing people to upgrade?

Jimexist:
@lucabergamini I have updated the PR, PTAL

Jimexist:
@lucabergamini to elaborate on my point, you can take a look at https://gist.github.com/Jimexist/745610edeffec18dda3ba96ffa243949

lucabergamini:
> @lucabergamini to elaborate on my point, you can take a look at https://gist.github.com/Jimexist/745610edeffec18dda3ba96ffa243949

I see...that's actually unexpected. One would think it would replace the cuda version with another cuda one.



> this actually makes sense. I guess you can drop the `< 1.6` and `< 0.7` for `torch` and `torchvision` without forcing people to upgrade?

Unless they re-install again from scratch it should not force them to install 1.6 I think

Jimexist:
> > @lucabergamini to elaborate on my point, you can take a look at https://gist.github.com/Jimexist/745610edeffec18dda3ba96ffa243949
> 
> I see...that's actually unexpected. One would think it would replace the cuda version with another cuda one.
> 
> > this actually makes sense. I guess you can drop the `< 1.6` and `< 0.7` for `torch` and `torchvision` without forcing people to upgrade?
> 
> Unless they re-install again from scratch it should not force them to install 1.6 I think

no I think the updated version will not force people to update but at the same time allow for newer version

lucabergamini (approved):
LGTM

Jimexist:
> LGTM

@lucabergamini I do not have the access to merge this, could you help with that? also I guess a new release is needed.

lucabergamini:
> @lucabergamini I do not have the access to merge this, could you help with that? also I guess a new release is needed.

Yeah I'll merge it ASAP and cut a new release. We want to include also other PRs in it so it won't happen right now :)

PR #140, opened 2020-09-09
Speedup semantic rasterizer

pestipeti:
I achieved ~ 7-8% speedup with these modifications.

### DIFF
--- a/l5kit/l5kit/data/map_api.py
+++ b/l5kit/l5kit/data/map_api.py
@@ -133,6 +133,7 @@ class MapAPI:
         )
 
         xyz = np.vstack((xyz_left, np.flip(xyz_right, 0)))
+        xyz[:, -1] = 1.0
 
         return {"xyz": xyz.T}
 
@@ -176,6 +177,7 @@ class MapAPI:
             traffic_element.geo_frame,
         )
 
+        xyz[:, -1] = 1.0
         return {"xyz": xyz.T}
 
     def is_traffic_face_colour(self, element_id: str, colour: str) -> bool:
--- a/l5kit/l5kit/data/map_api.py
+++ b/l5kit/l5kit/data/map_api.py
@@ -134,7 +134,7 @@ class MapAPI:
 
         xyz = np.vstack((xyz_left, np.flip(xyz_right, 0)))
 
-        return {"xyz_left": xyz_left, "xyz_right": xyz_right, "xyz": xyz}
+        return {"xyz_left": xyz_left, "xyz_right": xyz_right, "xyz": xyz.T}
 
     @staticmethod
     @no_type_check
@@ -176,7 +176,7 @@ class MapAPI:
             traffic_element.geo_frame,
         )
 
-        return {"xyz": xyz}
+        return {"xyz": xyz.T}
 
     def is_traffic_face_colour(self, element_id: str, colour: str) -> bool:
         """
--- a/l5kit/l5kit/data/map_api.py
+++ b/l5kit/l5kit/data/map_api.py
@@ -110,7 +110,7 @@ class MapAPI:
             element_id (str): lane element id
 
         Returns:
-            dict: a dict with the two boundaries coordinates as (Nx3) XYZ arrays
+            dict: a dict with the boundaries coordinates as (3xN) XYZ arrays
         """
         element = self[element_id]
         assert self.is_lane(element)
@@ -164,7 +164,7 @@ class MapAPI:
             element_id (str): crosswalk element id
 
         Returns:
-            dict: a dict with the polygon coordinates as an (Nx3) XYZ array
+            dict: a dict with the polygon coordinates as an (3xN) XYZ array
         """
         element = self[element_id]
         assert self.is_crosswalk(element)
--- a/l5kit/l5kit/data/map_api.py
+++ b/l5kit/l5kit/data/map_api.py
@@ -132,7 +132,9 @@ class MapAPI:
             lane.geo_frame,
         )
 
-        return {"xyz_left": xyz_left, "xyz_right": xyz_right}
+        xyz = np.vstack((xyz_left, np.flip(xyz_right, 0)))
+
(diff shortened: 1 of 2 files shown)


l5kit/l5kit/data/map_api.py
```diff
@@ -174,7 +177,8 @@ def get_crosswalk_coords(self, element_id: str) -> dict:
             traffic_element.geo_frame,
         )
 
-        return {"xyz": xyz}
+        xyz[:, -1] = 1.0
```
lucabergamini, line 180: why is this required? We ignore the z coordinates in the following so this shouldn't make any difference right?

l5kit/l5kit/data/map_api.py
```diff
@@ -174,7 +177,8 @@ def get_crosswalk_coords(self, element_id: str) -> dict:
...
+        return {"xyz": xyz.T}
```
lucabergamini, line 181: I can see the point of caching this to avoid stack at runtime, but we may still want to include either the two lanes or at least the length of the first, so that we can always unpack the two apart (I'm thinking about centre line support right now)

l5kit/l5kit/rasterization/semantic_rasterizer.py
```diff
@@ -176,12 +176,10 @@ def render_semantic_map(
 
             # get image coords
             lane_coords = self.proto_API.get_lane_coords(self.bounds_info["lanes"]["ids"][idx])
-            xy_left = cv2_subpixel(transform_points(lane_coords["xyz_left"][:, :2], world_to_image_space))
-            xy_right = cv2_subpixel(transform_points(lane_coords["xyz_right"][:, :2], world_to_image_space))
-            lanes_area = np.vstack((xy_left, np.flip(xy_right, 0)))  # start->end left then end->start right
+            lanes_xy = cv2_subpixel(world_to_image_space.dot(lane_coords["xyz"]).T[:, :2])
```
lucabergamini, line 179: Why not using `transform` or `transform_transpose` functions here?

lucabergamini (commented):
Amazing work! minors, you can either pick them up yourself or I can also do them, just let me know :)

l5kit/l5kit/data/map_api.py
pestipeti: In the current implementation, first you cut the z-coordinate `lane_coords["xyz_right"][:, :2]` and later in the transform method you stack it back `np.vstack((points[:num_dims, :], np.ones(points.shape[1])))`. With this we can save the one cut, the np.ones creation, the stack.

l5kit/l5kit/data/map_api.py
pestipeti: The cheapest/easiest solution if you add the length of the first line. I think `map_api` should handle the centerline calculation as well (cachable).

l5kit/l5kit/rasterization/semantic_rasterizer.py
pestipeti: It seemed a bit complicated. Transposes, vstack, np.ones, cut z-axis, etc. I did not want to break other parts of the code by modifying those methods.

(Feel free to correct anything you want.)

l5kit/l5kit/data/map_api.py
lucabergamini: Another way of doing this would be to only use the first 2x2 of the matrix (only XY) in the semantic rasterizer. My issue with setting 1 here is that we're removing information in a very hidden part of the code, which may render debugging problematic in the future (also considering there is a cache system in between)

l5kit/l5kit/data/map_api.py
lucabergamini: yeah, I agree on that and I'm fine with having the length included

l5kit/l5kit/rasterization/semantic_rasterizer.py
lucabergamini: Yeah you're perfectly right about that. Let me see if we can also handle this case there (same len for matrix and points)

l5kit/l5kit/data/map_api.py
pestipeti: I see. You can move it to the semantic rasterizer method; before the dot product calls.

l5kit/l5kit/data/map_api.py
pestipeti: One note: I changed the returned `xyz` format to `xyz.T`.

l5kit/l5kit/rasterization/semantic_rasterizer.py
pestipeti: I did these modifications only to save as many CPU cycles as I possibly can. I did not consider long term usability, I only care about speeding up for the competition, so it is fine if you drop all of or part of this PR.

l5kit/l5kit/data/map_api.py
lucabergamini: I like this idea also :) Today was quite busy, but I should be able to work on this tomorrow (hopefully)

pascal-pfeiffer:
render_semantic_map is the bottleneck when using a low count of history_frames.
This PR speeds up the render_semantic_map function by ~200%.
Even for higher count of history_frames this change is still significant and I suggest to accept it.

lucabergamini:
> This PR speeds up the render_semantic_map function by ~200%.

did you compare this against #196?

pascal-pfeiffer:
> did you compare this against #196?

thanks for pointing to that PR. It does even better once the caching is done:

52 ms ± 4.77 ms per loop (mean ± std. dev. of 7 runs, 10 loops each) no changes
24.8 ms ± 777 µs per loop (mean ± std. dev. of 7 runs, 10 loops each) this PR #140

32.9 ms ± 17.2 ms per loop (mean ± std. dev. of 7 runs, 10 loops each) PR #196 
The slowest run took 4.23 times longer than the fastest. This could mean that an intermediate result is being cached.

after caching:
17.5 ms ± 880 µs per loop (mean ± std. dev. of 7 runs, 100 loops each) PR #196

Issue #137, opened 2020-09-06
bug in agent sampling?


jmercat:
I found this line in the file l5kit/l5kit/l5kit/sampling/agent_sampling.py line 193, "agents" is both the iterated element and the list on which the loop is done. I'm not sure what this means but this should probably be corrected.

`for i, (frame, agents) in enumerate(zip(frames, agents)):`

lucabergamini:
Interesting! It's actually weird everything has been working properly, I'll dig into into it!

lucabergamini:
So it turned out this is perfectly legal in Python (see #141 ). Still, we're addressing it for clarity :) I'll change the label to `enhancement`.

lucabergamini:
#141 has been merged into master, closing this now :)

PR #141, opened 2020-09-09
rename same-name loop variable

lucabergamini:
We are using the same name for the looped variable and the item, see #137 

A minimal example of the issue follows:
```python
els = [1,2,3]
for els in els:
    print(els)
```

Note this is perfectly legal in Python ( my guess is that the iterable is store as a ref so reassigning the name doesn't affect it) and produces
```
1
2
3
```
as expected. Still, we should change the name for clarity

### DIFF
--- a/l5kit/l5kit/sampling/agent_sampling.py
+++ b/l5kit/l5kit/sampling/agent_sampling.py
@@ -190,14 +190,14 @@ def _create_targets_for_deep_prediction(
     yaws_offset = np.zeros((num_frames, 1), dtype=np.float32)
     availability = np.zeros((num_frames,), dtype=np.float32)
 
-    for i, (frame, agents) in enumerate(zip(frames, agents)):
+    for i, (frame, frame_agents) in enumerate(zip(frames, agents)):
         if selected_track_id is None:
             agent_centroid = frame["ego_translation"][:2]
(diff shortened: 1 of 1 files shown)


l5kit/l5kit/sampling/agent_sampling.py
```diff
@@ -190,14 +190,14 @@ def _create_targets_for_deep_prediction(
     yaws_offset = np.zeros((num_frames, 1), dtype=np.float32)
     availability = np.zeros((num_frames,), dtype=np.float32)
 
-    for i, (frame, agents) in enumerate(zip(frames, agents)):
+    for i, (frame, frame_agents) in enumerate(zip(frames, agents)):
```
oliver-scheel, line 193: Minor: how about agent?

oliver-scheel: Spoke too soon, seeing the contents of agents now. I just liked the notation frames -> frame, agents-> agent. frame_agents is a bit different, but well, minor. and would not know how to improve.

oliver-scheel (approved):
LGTM

l5kit/l5kit/sampling/agent_sampling.py
lucabergamini: yeah..naming is not our top skill, let's put it that way xD

PR #145, opened 2020-09-10
[CI] run github ci on various torch versions

Jimexist:
I added in a previous PR to allow torch 1.6, but I think we can allow GitHub to capture this various versions in a matrix.

### DIFF
--- a/.github/workflows/l5kit.yml
+++ b/.github/workflows/l5kit.yml
@@ -2,7 +2,7 @@ name: L5Kit-tests
 
 on:
   push:
-    branches: [ master ]
+    branches: [master]
   pull_request:
 
 jobs:
@@ -11,6 +11,7 @@ jobs:
     strategy:
       matrix:
         python-version: [3.6, 3.7, 3.8]
+        torch-deps: [15, 16]
 
     steps:
     - uses: actions/checkout@v2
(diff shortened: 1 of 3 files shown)


Jimexist:
Guess I'll have to take into account torchvision versions as well

lucabergamini:
mmm this workflow actually looks invalid to GitHub (`torch-deps` is not recognised as a valid command). I don't know if this is something we can handle there actually.

Jimexist:
> mmm this workflow actually looks invalid to GitHub (`torch-deps` is not recognised as a valid command). I don't know if this is something we can handle there actually.

yeah I took a different approach and updated this PR

Jimexist:
@lucabergamini what do you think?

lucabergamini:
> @lucabergamini what do you think?

Sorry for the late reply. The issue is that this will conflict with #126  as I'm not using requirements there for deterministic builds. I'm considering if this is something we actually want to test at this point...

Jimexist:
> > @lucabergamini what do you think?
> 
> Sorry for the late reply. The issue is that this will conflict with #126 as I'm not using requirements there for deterministic builds. I'm considering if this is something we actually want to test at this point...

fair point, let me close this.

PR #153, opened 2020-09-25
Lberg/refactor transform functions

lucabergamini:
Avoid Stacking and slicing. Add better docstring. Drop 4D points support.

### DIFF
--- a/l5kit/l5kit/geometry/__init__.py
+++ b/l5kit/l5kit/geometry/__init__.py
@@ -8,7 +8,6 @@ from .transform import (
     rotation33_as_yaw,
     transform_point,
     transform_points,
-    transform_points_transposed,
     world_to_image_pixels_matrix,
     yaw_as_rotation33,
 )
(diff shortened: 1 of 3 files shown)


yye-lyft:
I implemented something similar using Tensor which avoid `transpose` the points, I would like to get ride of `transform_points_transposed `: https://github.com/lyft/avresearch/blob/yye/refactor-transform-libs/avresearch/data/transforms.py#L29

lucabergamini:
> I would like to get rid of `transform_points_transposed`

Take a look at the function now @yye-lyft. If you're fine with that we can remove the transposed version

yye-lyft:
> > I would like to get rid of `transform_points_transposed`
> 
> Take a look at the function now @yye-lyft. If you're fine with that we can remove the transposed version

Yeah, let's remove the transposed version

yye-lyft (dismissed):
LGTM, nice work!

yye-lyft (approved):
LGTM

PR #162, opened 2020-10-05
use wider range of torch and torchvision version

Jimexist:
I made same changes in `setup.py` but not in the new `Pipfile`


lucabergamini:
> I made same changes in `setup.py` but not in the new `Pipfile`

Makes sense. The `setup.py` is still tested so we should know if something  breaks

lucabergamini (approved):
LGTM

yawei-ye (approved):
LGTM

PR #163, opened 2020-10-05
fix metrics.py assertion error messages

Jimexist:


l5kit/l5kit/evaluation/metrics.py
```diff
@@ -21,12 +21,13 @@ def _assert_shapes(ground_truth: np.ndarray, pred: np.ndarray, confidences: np.n
...
-    )
-    assert confidences.shape == (num_modes,), f"expected 1D (Modes) array for gt, got {confidences.shape}"
+    assert ground_truth.shape == (
+        future_len,
+        num_coords,
+    ), f"expected 2D (Time x Coords) array for gt, got {ground_truth.shape}"
+    assert confidences.shape == (num_modes,), f"expected 1D (Modes) array for confidences, got {confidences.shape}"
     assert np.allclose(np.sum(confidences), 1), "confidences should sum to 1"
-    assert avails.shape == (future_len,), f"expected 1D (Time) array for gt, got {avails.shape}"
+    assert avails.shape == (future_len,), f"expected 1D (Time) array for avails, got {avails.shape}"
```
lucabergamini, line 30: ah good catch :)

l5kit/l5kit/evaluation/metrics.py
```diff
@@ -21,12 +21,13 @@ def _assert_shapes(ground_truth: np.ndarray, pred: np.ndarray, confidences: np.n
     assert len(pred.shape) == 3, f"expected 3D (MxTxC) array for pred, got {pred.shape}"
     num_modes, future_len, num_coords = pred.shape
 
-    assert ground_truth.shape == (future_len, num_coords), (
-        f"expected 2D (Time x Coords) array for gt, " f"got {ground_truth.shape}"
-    )
-    assert confidences.shape == (num_modes,), f"expected 1D (Modes) array for gt, got {confidences.shape}"
+    assert ground_truth.shape == (
+        future_len,
+        num_coords,
```
lucabergamini, line 26: ugh a trailing comma here is making things even more hard to read.
I can't decide if we should have it here (given we don't plan to expand that tuple) or not (but then we have it for other tuples in the codebase)

lucabergamini (approved):
LGTM

l5kit/l5kit/evaluation/metrics.py
Jimexist: It exists to make linter happy.

in fact I hold rather a radical point of view of code formatting (e.g. I'd prefer [black](https://github.com/psf/black) for python, [google java format](https://github.com/google/google-java-format) for java, [prettier](https://prettier.io) for JavaScript, and they all have one thing in common: opinionated and guides all contributes away from spending another minute on code formatting since it isn't worthy of the time)

lucabergamini: yeah, in this specific case the trailing comma is not required (PEP8 is strict about it only for sequences with just one element). It's not something that black or flake will pick up. I'm fine with the change anyway :)

Jimexist: yes i think black would enforce it. https://black.readthedocs.io/en/stable/the_black_code_style.html#trailing-commas

to me it makes more sense because it makes git diff cleaner, and what's more it makes it less error prone when e.g. changing from two to one element tuple where the trailing comma suddenly is meaningful... anyway i vowed not to dive too much into formatting styles ;-) i'm fine whenever linter is happy

PR #170, opened 2020-10-08
allow users to limit the number of history frames to rasterize

Jimexist:
for now the configurations for rasterizing history frames and for returning within the batch data are tightly coupled, if one wants to use more historical context and do not need those history frames to be rasterized as boxes, an exit hatch can be handy.

### DIFF
--- a/examples/agent_motion_prediction/agent_motion_config.yaml
+++ b/examples/agent_motion_prediction/agent_motion_config.yaml
@@ -43,6 +43,10 @@ raster_params:
   # whether to completely disable traffic light faces in the semantic rasterizer
   disable_traffic_light_faces: False
 
+  # if you want a larger number of history frames but would not want to rasterize them all
+  # this will override .model_params.history_num_frames
+  # history_num_frames_to_rasterize: 0
+
 ###################
 ## Data loader options
 train_data_loader:
--- a/l5kit/l5kit/rasterization/rasterizer_builder.py
+++ b/l5kit/l5kit/rasterization/rasterizer_builder.py
@@ -131,6 +131,19 @@ def build_rasterizer(cfg: dict, data_manager: DataManager) -> Rasterizer:
     filter_agents_threshold = raster_cfg["filter_agents_threshold"]
     history_num_frames = cfg["model_params"]["history_num_frames"]
 
+    if "history_num_frames_to_rasterize" in raster_cfg:
+        _override_val = raster_cfg["history_num_frames_to_rasterize"]
+        assert 0 <= _override_val <= history_num_frames, (
+            f".raster_params.history_num_frames_to_rasterize ({_override_val}) should be in"
+            " the range of [0, history_num_frames]"
+        )
+        print(
+            "You have overridden history_num_frames in the rasterizer,",
+            f"which means you will have {history_num_frames} frames of raw data but only",
+            f"{_override_val} rasterized as images",
+        )
+        history_num_frames = _override_val
+
     if map_type in ["py_satellite", "satellite_debug"]:
         sat_image = _load_satellite_map(raster_cfg["satellite_map_key"], data_manager)
 
--- a/l5kit/l5kit/tests/dataset/dataset_test.py
+++ b/l5kit/l5kit/tests/dataset/dataset_test.py
@@ -1,4 +1,3 @@
-from copy import deepcopy
 from typing import Callable
 
 import numpy as np
@@ -42,7 +41,6 @@ def check_torch_loading(dataset: Dataset) -> None:
 def test_dataset_rasterizer(
     rast_name: str, dataset_cls: Callable, zarr_dataset: ChunkedDataset, dmg: LocalDataManager, cfg: dict
 ) -> None:
-    cfg = deepcopy(cfg)
     cfg["raster_params"]["map_type"] = rast_name
     rasterizer = build_rasterizer(cfg, dmg)
 
@@ -56,7 +54,6 @@ def test_dataset_rasterizer(
 def test_dataset_rasterizer_with_override(
     rast_name: str, dataset_cls: Callable, zarr_dataset: ChunkedDataset, dmg: LocalDataManager, cfg: dict
 ) -> None:
-    cfg = deepcopy(cfg)
     cfg["raster_params"]["map_type"] = rast_name
     cfg["model_params"]["history_num_frames"] = 1
     cfg["raster_params"]["history_num_frames_to_rasterize"] = 0
--- a/l5kit/l5kit/tests/dataset/dataset_test.py
+++ b/l5kit/l5kit/tests/dataset/dataset_test.py
@@ -1,3 +1,4 @@
+from copy import deepcopy
 from typing import Callable
 
 import numpy as np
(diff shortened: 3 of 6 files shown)


lucabergamini:
> if one wants to use more historical context and do not need those history frames to be rasterized as boxes

mmm, can you provide me an example of this? I'm not sure I'm getting the current issue

Jimexist:
> > if one wants to use more historical context and do not need those history frames to be rasterized as boxes
> 
> mmm, can you provide me an example of this? I'm not sure I'm getting the current issue

For each item in `AgentDataset` it is a dict of this shape:

```python
{
  "image": image_vector,
  "history_positions: history_positions_vector,
  "history_yaws": history_yaws_vector,
  # amongst other values
}
```

where [`image_vector`](https://github.com/lyft/l5kit/blob/03eb9e037d23940a134e27c9f124021e18982020/l5kit/l5kit/dataset/ego.py#L112) is of size `(batch, 3 + 2 * (history_num_frames + 1), width, height)` and [`history_positions_vector`](https://github.com/lyft/l5kit/blob/03eb9e037d23940a134e27c9f124021e18982020/l5kit/l5kit/dataset/ego.py#L116) is of size `(batch, history_num_frames + 1)`. That means if I only want history_positions_vector to have `size[1] == 99` while I do not need all of these history frames to be rasterized, then I have no choice but to rasterize just to discard them.

pascal-pfeiffer:
> mmm, can you provide me an example of this? I'm not sure I'm getting the current issue

I can definetly see a usage here. Imagine someone who wants to use the rasterizer soley for the current timestep. The other information about past positions (Agent vehicle, no other vehicles) will still be provided by the "history_positions" and can be used for e.g. a kalman filter.

It depends on the actual model, but it would open more possibilities by keeping the CPU load managable.

lucabergamini:
> It depends on the actual model, but it would open more possibilities by keeping the CPU load managable.

I see your point, but we're paying a high price in complexity here. 
To get the default behaviour you now have to delete a key from the config -> easy to forget.
If you're looking at the config to build your network now you need to check this param, otherwise you get an error at runtime.

It's something worth having if we've a lot of users asking for it (and we must be very careful with the implementation). Right now, it seems to me we don't have this huge demand for this feature..

lucabergamini:
which doesn't mean we won't take this PR btw, but I have to check a couple of things to keep the code as simple as possible :)

lucabergamini:
> That means if I only want history_positions_vector to have `size[1] == 99` while I do not need all of these history frames to be rasterized, then I have no choice but to rasterize just to discard them.

Just to be super clear on this, you want **some** history in raster but not all of it (because if you're not interested in raster at all `stub_debug` will do in that case). Am I correct?

pascal-pfeiffer:
> > That means if I only want history_positions_vector to have `size[1] == 99` while I do not need all of these history frames to be rasterized, then I have no choice but to rasterize just to discard them.
> 
> Just to be super clear on this, you want **some** history in raster but not all of it (because if you're not interested in raster at all `stub_debug` will do in that case). Am I correct?

This is exactly how I understand it. It may help when combining CNN models with e.g. traditional Kalman filters.

examples/agent_motion_prediction/agent_motion_config.yaml
```diff
@@ -43,6 +43,10 @@ raster_params:
   # whether to completely disable traffic light faces in the semantic rasterizer
   disable_traffic_light_faces: False
 
+  # if you want a larger number of history frames but would not want to rasterize them all
+  # this will override .model_params.history_num_frames
+  rasterize_history_num_frames_override: 0
```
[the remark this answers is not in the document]
Jimexist: maybe `history_num_frames_to_rasterize`?

lucabergamini:
> This is exactly how I understand it. It may help when combining CNN models with e.g. traditional Kalman filters.

Agree, let me come up with the best design I can think for this feature then

l5kit/l5kit/rasterization/rasterizer_builder.py
```diff
@@ -131,6 +131,18 @@ def build_rasterizer(cfg: dict, data_manager: DataManager) -> Rasterizer:
     filter_agents_threshold = raster_cfg["filter_agents_threshold"]
     history_num_frames = cfg["model_params"]["history_num_frames"]
 
+    if "history_num_frames_to_rasterize" in raster_cfg:
+        _override_val = raster_cfg["history_num_frames_to_rasterize"]
```
lucabergamini, line 135: I would put a warning here
"you're overwriting history_num_frames in the rasteriser only, image and history_positions will have different lengths at runtime"

This way we have at least a warning when the forward will fail xD

l5kit/l5kit/rasterization/rasterizer_builder.py
```diff
@@ -131,6 +131,18 @@ def build_rasterizer(cfg: dict, data_manager: DataManager) -> Rasterizer:
...
+        assert isinstance(
+            _override_val, int
+        ), f".raster_params.history_num_frames_to_rasterize ({_override_val}) must be an int"
+        assert 0 <= _override_val <= history_num_frames, (
+            f".raster_params.history_num_frames_to_rasterize ({_override_val}) should be in"
+            " the range of [0, history_num_frames]"
+        )
+        history_num_frames = _override_val
+        del _override_val
```
lucabergamini, line 144: no need to `del` I think :)

l5kit/l5kit/tests/rasterization/box_rasterizer_test.py
```diff
@@ -87,3 +89,27 @@ def test_out_shape(hist_data: tuple, dmg: LocalDataManager, cfg: dict) -> None:
...
+
+@pytest.mark.parametrize("lengths", [(5, 5), (3, 3), (5, 0), (0, 0)])
+def test_out_shape_override(hist_data: tuple, dmg: LocalDataManager, cfg: dict, lengths: Tuple[int, int]) -> None:
+    hist_length, real_length = lengths
+    cfg["raster_params"]["map_type"] = "box_debug"
+    cfg["raster_params"]["history_num_frames_to_rasterize"] = real_length
+    cfg["model_params"]["history_num_frames"] = hist_length
+    rasterizer = build_rasterizer(cfg, dmg)
+    out = rasterizer.rasterize(hist_data[0][: hist_length + 1], hist_data[1][: hist_length + 1], [])
+    assert out.shape == (224, 224, (real_length + 1) * 2)
```
lucabergamini, line 102: can we have a test where a `EgoDataset` is built and `history_positions` and `image` have different lengths?

lucabergamini (commented):
minors, but I think the direction is the right one :)

l5kit/l5kit/tests/dataset/dataset_test.py
```diff
@@ -31,6 +42,24 @@ def check_torch_loading(dataset: Dataset) -> None:
 def test_dataset_rasterizer(
     rast_name: str, dataset_cls: Callable, zarr_dataset: ChunkedDataset, dmg: LocalDataManager, cfg: dict
 ) -> None:
+    cfg = deepcopy(cfg)
+    cfg["raster_params"]["map_type"] = rast_name
```
Jimexist, line 46: @lucabergamini  a bug that `rast_name` was not used

lucabergamini: good catch!

lucabergamini: I'd say you don't need the deepcopy as the fixture's scope is set to `function`

Jimexist:
@lucabergamini comments address, please take another look, thanks!

PR #172, opened 2020-10-10
Removed OS generated file .DS_Store

pascal-pfeiffer:
- Removed OS generated file .DS_Store in folder images
 -Modified .gitignore


lucabergamini (approved):
LGTM!

PR #173, opened 2020-10-10
fix sem rast field naming

Jimexist:
the semantic rasterizer has wrong field naming possibly from copy and paste.


pascal-pfeiffer (changes requested):
Also change the `im_out_sat` to `im_out_sem` when you are at it ;)

pascal-pfeiffer (approved):

lucabergamini (approved):
it always surprises me how many naming errors I can put in my code :)

PR #174, opened 2020-10-10
Combine sem/sat + box rast into a map and box rast

Jimexist:
the sem/sat + box rasterizers have lots of duplicated code, refactor them so that they are minimal and can share code.


Jimexist:
i guess with this we no longer need to have https://github.com/lyft/l5kit/pull/173

PR #175, opened 2020-10-10
upgrade and use black 20

Jimexist:
seems that:
1. black now handles docstring more properly
2. it also folds params more aggressively

### DIFF
--- a/l5kit/Pipfile
+++ b/l5kit/Pipfile
@@ -13,7 +13,7 @@ sphinx = "*"
 sphinx-rtd-theme = "*"
 recommonmark = "*"
 pre-commit = "*"
-black = "==19.10b0"
+black = "==20.8b1"
 importlib-metadata = "*"  # required for python 3.8
 
 [packages]
--- a/l5kit/l5kit/data/zarr_dataset.py
+++ b/l5kit/l5kit/data/zarr_dataset.py
@@ -47,7 +47,7 @@ AGENT_DTYPE = [
 TL_FACE_DTYPE = [
     ("face_id", "<U16"),
     ("traffic_light_id", "<U16"),
-    ("traffic_light_face_status", np.float32, (len(TL_FACE_LABELS,))),
+    ("traffic_light_face_status", np.float32, (len(TL_FACE_LABELS),)),
 ]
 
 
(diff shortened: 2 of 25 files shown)


Jimexist:
maybe not worth it...

pascal-pfeiffer:
I see some issues in the Docstrings where `Args` and `Returns` are sometimes on different indentation levels.

lucabergamini:
> * black now handles docstring more properly
> * it also folds params more aggressively

I'm fine with the first point (I like it!) but the second seems **really** aggressive. Is this the new default?

Jimexist:
> > * black now handles docstring more properly
> > * it also folds params more aggressively
> 
> I'm fine with the first point (I like it!) but the second seems **really** aggressive. Is this the new default?

I think it's something with explicit trailing comma, if you remove them, black will no longer force line folding.

l5kit/l5kit/data/zarr_dataset.py
```diff
@@ -47,7 +47,7 @@
 TL_FACE_DTYPE = [
     ("face_id", "<U16"),
     ("traffic_light_id", "<U16"),
-    ("traffic_light_face_status", np.float32, (len(TL_FACE_LABELS,))),
+    ("traffic_light_face_status", np.float32, (len(TL_FACE_LABELS),)),
```
Jimexist, line 50: i thought this should be a bug @lucabergamini

l5kit/Pipfile
```diff
@@ -8,12 +8,12 @@ pytest = "*"
 mypy = "*"
 pytest-cov = "*"
 flake8 = "*"
-isort = "*"
+isort = "==5.6.2"
```
lucabergamini, line 11: why pinning this?

lucabergamini (commented):
minor :)

l5kit/Pipfile
Jimexist: like black, these things change and might silently break CI.

lucabergamini: yes please, we have tests in place just to catch those issues (the ones without the lock) :)

lucabergamini (approved):
LGTM

PR #176, opened 2020-10-10
allow rasterizer and returned image to be None

Jimexist:
I think currently if you use None rasterizer it'll throw.

### DIFF
--- a/l5kit/l5kit/dataset/ego.py
+++ b/l5kit/l5kit/dataset/ego.py
@@ -79,8 +79,9 @@ None if not desired
             state_index (int): a relative frame index in the scene
             track_id (Optional[int]): the agent to rasterize or None for the AV
         Returns:
-            dict: the rasterised image, the target trajectory (position and yaw) along with their availability,
-            the 2D matrix to center that agent, the agent track (-1 if ego) and the timestamp
+            dict: the rasterised image in (Cx0x1) if the rast is not None, the target trajectory
+            (position and yaw) along with their availability, the 2D matrix to center that agent,
+            the agent track (-1 if ego) and the timestamp
 
         """
         frames = self.dataset.frames[get_frames_slice_from_scenes(self.dataset.scenes[scene_index])]
--- a/l5kit/l5kit/dataset/ego.py
+++ b/l5kit/l5kit/dataset/ego.py
@@ -96,8 +96,12 @@ None if not desired
                 stacklevel=2,
             )
         data = self.sample_function(state_index, frames, self.dataset.agents, tl_faces, track_id)
-        # 0,1,C -> C,0,1
-        image = data["image"].transpose(2, 0, 1)
+
+        # when rast is None, image could be None
+        image = data["image"]
+        if image is not None:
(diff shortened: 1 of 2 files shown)


lucabergamini:
> I think currently if you use None rasterizer it'll throw.

yeah, the case with no rasterizer is currently not handled. I'm not sure this PR fixes it though. Can This be placed into a `DataLoader`?  Last time I've checked `None` was not something the default collate function was able to handle

Jimexist:
> > I think currently if you use None rasterizer it'll throw.
> 
> yeah, the case with no rasterizer is currently not handled. I'm not sure this PR fixes it though. Can This be placed into a `DataLoader`? Last time I've checked `None` was not something the default collate function was able to handle

@lucabergamini updated, also if it is designed not to return none then I guess we can just use stub rasterizer in the absence of a rasterizer?

lucabergamini:
> @lucabergamini updated, also if it is designed not to return none then I guess we can just use stub rasterizer in the absence of a rasterizer?

We have to make a choice:
1. when you pass None you don't get an image at all in the dict;
2. when you pass None a StubRasterizer is created and you get a black image;

I can see pros and cons in both, what do you think is better and why @Jimexist ?

1. nice because semantically you don't want an image, bad for consistency (but maybe only in tests)
2. nice for consistency but the field actually makes no sense

Another option would be to require a rasterizer always

Jimexist:
> > @lucabergamini updated, also if it is designed not to return none then I guess we can just use stub rasterizer in the absence of a rasterizer?
> 
> We have to make a choice:
> 
> 1. when you pass None you don't get an image at all in the dict;
> 2. when you pass None a StubRasterizer is created and you get a black image;
> 
> I can see pros and cons in both, what do you think is better and why @Jimexist ?
> 
> 1. nice because semantically you don't want an image, bad for consistency (but maybe only in tests)
> 2. nice for consistency but the field actually makes no sense
> 
> Another option would be to require a rasterizer always

I'd prefer 1st option because users who explicitly pass in `None` for rasterizer would expect the image to be not there or at least can handle the discrepancy of returned items' keys; while if we silently use `StubRasterizer` there's a chance of misuse.

If we had to go with 2nd option, I'd suggest putting `torch.tensor(0.)` or even `torch.tensor(np.nan)` just so that it's not used by chance (and also saves memory).

lucabergamini:
> I'd prefer 1st option because users who explicitly pass in `None` for rasterizer would expect the image to be not there or at least can handle the discrepancy of returned items' keys; while if we silently use `StubRasterizer` there's a chance of misuse

I agree, and as far as I can see it's what you've already implemented, so I'll approve it :)

lucabergamini (approved):
LGTM

lucabergamini:
> I'd prefer 1st option because users who explicitly pass in `None` for rasterizer would expect the image to be not there or at least can handle the discrepancy of returned items' keys; while if we silently use `StubRasterizer` there's a chance of misuse

I agree, and as far as I can see it's what you've already implemented, so I'll approve it :)

PR #177, opened 2020-10-10
Transformation speedup

pascal-pfeiffer:
My tests suggest that the modifications give a slight speed increase when used for 2-20000 points. Speed increase is larger when the points array is larger. For tests I purposly I used a continuous and non-continuous arrays with very similar results.

N = 20000
old: 168 µs ± 5.91 µs per loop (mean ± std. dev. of 7 runs, 10000 loops each)
new: 159 µs ± 4.89 µs per loop (mean ± std. dev. of 7 runs, 10000 loops each)

N = 20
old: 9.5 µs ± 139 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)
new: 9.31 µs ± 173 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)

I also modified the test for the above function as I found ways to bypass the test with "easy" solutions.


l5kit/l5kit/geometry/transform.py
```diff
@@ -92,14 +92,11 @@ def transform_points(points: np.ndarray, transf_matrix: np.ndarray) -> np.ndarra
...
-        raise AssertionError("Points input should be (N, 2) or (N, 3) shape, received {}".format(points.shape))
+        raise AssertionError(f"Points input should be (N, 2) or (N, 3) shape, received {points.shape}")
 
     assert points.shape[1] == transf_matrix.shape[1] - 1, "points dim should be one less than matrix dim"
 
-    num_dims = len(transf_matrix) - 1
-    transf_matrix = transf_matrix.T
-
-    return points @ transf_matrix[:num_dims, :num_dims] + transf_matrix[-1, :num_dims]
+    return (points @ transf_matrix.T[:-1, :-1]) + transf_matrix[:-1, -1]
```
lucabergamini, line 99: I don't really understand why this should be faster (maybe some memory alignment) but I'm fine with it :)

lucabergamini (approved):
LGTM

PR #179, opened 2020-10-11
add multi-agent dataset

Jimexist:
context explained in https://www.kaggle.com/etareduce/lyft-multi-agent-dataset-neighboring-by-distance


lucabergamini:
I'll close this as we're working on an efficient implementation of this concept internally :)

PR #189, opened 2020-11-03
compute estimated target velocity from target positions

yawei-ye:
# Summary

- compute estimated target velocities from target positions

### DIFF
--- a/l5kit/l5kit/dataset/ego.py
+++ b/l5kit/l5kit/dataset/ego.py
@@ -41,9 +41,6 @@ None if not desired
         self.rasterizer = rasterizer
 
         self.cumulative_sizes = self.dataset.scenes["frame_index_interval"][:, 1]
-        self.future_step_time = (
-            self.cfg["model_params"]["future_delta_time"] * self.cfg["model_params"]["future_step_size"]
-        )
 
         render_context = RenderContext(
             raster_size_px=np.array(cfg["raster_params"]["raster_size"]),
@@ -57,8 +54,10 @@ None if not desired
             render_context=render_context,
             history_num_frames=cfg["model_params"]["history_num_frames"],
             history_step_size=cfg["model_params"]["history_step_size"],
+            history_step_time=cfg["model_params"]["history_delta_time"] * cfg["model_params"]["history_step_size"],
             future_num_frames=cfg["model_params"]["future_num_frames"],
             future_step_size=cfg["model_params"]["future_step_size"],
+            future_step_time=cfg["model_params"]["future_delta_time"] * cfg["model_params"]["future_step_size"],
             filter_agents_threshold=cfg["raster_params"]["filter_agents_threshold"],
             rasterizer=rasterizer,
             perturbation=perturbation,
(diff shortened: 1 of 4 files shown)


yawei-ye:
meow @lucabergamini

l5kit/l5kit/dataset/ego.py
```diff
@@ -98,18 +101,38 @@ def get_frame(self, scene_index: int, state_index: int, track_id: Optional[int]
...
 
-        history_positions = np.array(data["history_positions"], dtype=np.float32)
-        history_yaws = np.array(data["history_yaws"], dtype=np.float32)
+        # compute estimated target velocities by finite differentiatin on target positions
+        curr_position = np.array([0.0, 0.0])
+        assert np.allclose(history_positions[0], curr_position), "current ego position should be normalized to origin"
+        # [future_num_frames, 2]
+        target_pos_diff = target_positions - np.append([curr_position], target_positions[:-1, :], axis=0)
+        # [future_num_frames,]
+        target_velocities = np.float32(np.linalg.norm(target_pos_diff / self.future_step_time, axis=1))
```
lucabergamini, line 120: this is velocity modules, are we sure we're not interested in the two XY components?

l5kit/l5kit/dataset/ego.py
```diff
@@ -98,18 +101,38 @@ def get_frame(self, scene_index: int, state_index: int, track_id: Optional[int]
             )
         data = self.sample_function(state_index, frames, self.dataset.agents, tl_faces, track_id)
 
+        # [history_num_frames + 1, 2]
```
lucabergamini, line 104: why are we computing this values here and not in the sampling function?

lucabergamini (commented):
Minors

l5kit/l5kit/dataset/ego.py
yawei-ye: can you help me to move it to sampling function?

l5kit/l5kit/dataset/ego.py
azhong-git: shall we return `target_velocities` in xy components and only return `speed` at time 0 (which uses np.linalg.norm on xy components)? the latter (scalalr `speed`) will be used in inference and the former (keeping both xy components) is helpful (e.g. when computing acceleration in x, y components)

l5kit/l5kit/dataset/ego.py
```diff
@@ -124,6 +147,7 @@ def get_frame(self, scene_index: int, state_index: int, track_id: Optional[int]
             "centroid": data["centroid"],
```
azhong-git, line 147: nit: shall we also return host id? in case we would like to reduce metrics among neighboring frames (e.g. diff of open loop velocities / acceleration among neighboring frames) which requires car id in addition to timestamp

l5kit/l5kit/dataset/ego.py
yawei-ye: moved to sampling function

yawei-ye:
meow @lucabergamini @azhong-git I addressed the review comments

l5kit/l5kit/dataset/ego.py
```diff
@@ -98,32 +100,45 @@ def get_frame(self, scene_index: int, state_index: int, track_id: Optional[int]
...
         data = self.sample_function(state_index, frames, self.dataset.agents, tl_faces, track_id)
 
-        target_positions = np.array(data["target_positions"], dtype=np.float32)
-        target_yaws = np.array(data["target_yaws"], dtype=np.float32)
-
+        # [history_num_frames + 1, 2]
         history_positions = np.array(data["history_positions"], dtype=np.float32)
+        # [history_num_frames + 1, 1]
         history_yaws = np.array(data["history_yaws"], dtype=np.float32)
+        # [future_num_frames + 1, 2]
```
lucabergamini, line 107: wrong comment?

l5kit/l5kit/dataset/ego.py
```diff
@@ -98,32 +100,45 @@ def get_frame(self, scene_index: int, state_index: int, track_id: Optional[int]
...
+        history_velocities = np.array(data["history_velocities"], dtype=np.float32)
+
+        # [future_num_frames, 2]
+        target_positions = np.array(data["target_positions"], dtype=np.float32)
+        # [future_num_frames, 1]
+        target_yaws = np.array(data["target_yaws"], dtype=np.float32)
+        # [future_num_frames, 2]
+        target_velocities = np.array(data["target_velocities"], dtype=np.float32)
 
+        host_id = (self.dataset.scenes[scene_index]["host"],)
```
lucabergamini, line 117: you can also remove the tuple. Torch will give you a list of strings.

What I'm worried about is what will happen when you try to move the entire dict in cuda though

l5kit/l5kit/dataset/ego.py
```diff
@@ -54,8 +54,10 @@ def __init__(
             render_context=render_context,
             history_num_frames=cfg["model_params"]["history_num_frames"],
             history_step_size=cfg["model_params"]["history_step_size"],
+            history_step_time=cfg["model_params"]["history_delta_time"] * cfg["model_params"]["history_step_size"],
```
lucabergamini, line 57: I'm fine in leaving it as it is for now, but we should really decide if we want all this flexibility. I think we should just remove step_size all together honestly..

lucabergamini (dismissed):
some minors but LGTM :)

l5kit/l5kit/dataset/ego.py
yawei-ye: good catch

l5kit/l5kit/dataset/ego.py
yawei-ye: hmm, it is caused by autoformatting, addressed

lucabergamini (approved):

PR #193, opened 2020-11-06
Make EgoDataset get_frame transparent

lucabergamini:
avoid redeclaring dict, so that all keys are preserved.
This allow to replace the default sampling function and extend
functionality of the function without subclassing


yawei-ye (approved):
LGTM!

anasrferreira (approved):
Thanks for the change

PR #194, opened 2020-11-09
Disable nolock on Windows

lucabergamini:
Disable BLOSC_NOLOCK on Windows, which may cause OOM errors.
However, add a warning as this may cause issue when writing the mask


yawei-ye (approved):
LGTM

PR #196, opened 2020-11-10
[AVR-761] Lberg/refactor map api

lucabergamini:
Refactor of MapAPI:

- move bounds computation inside the mapAPI. This makes the mapAPI object more independent;
- perform interpolation of lanes. This allows to speed up transformations between spaces


anasrferreira (dismissed):
I am guessing that the goal of exposing this is that we can select the elements of interest by their bounds, right?
Though for the geometry itself we will want to use get_lane_coords and get_crosswalk_coords
Is this right @lucabergamini

lucabergamini:
> I am guessing that the goal of exposing this is that we can select the elements of interest by their bounds, right?
> Though for the geometry itself we will want to use get_lane_coords and get_crosswalk_coords

Exactly :)

lucabergamini:
PTAL @anasrferreira , I've added a function to sample from the lane coords

l5kit/l5kit/rasterization/semantic_rasterizer.py
```diff
@@ -15,7 +15,7 @@
 CV2_SHIFT_VALUE = 2 ** CV2_SHIFT
 
 
-def elements_within_bounds(center: np.ndarray, bounds: np.ndarray, half_extent: float) -> np.ndarray:
+def indices_in_bounds(center: np.ndarray, bounds: np.ndarray, half_extent: float) -> np.ndarray:
```
anasrferreira, line 18: Should this one also get moved to map_api?

l5kit/l5kit/data/map_api.py
```diff
@@ -134,6 +142,53 @@ def get_lane_coords(self, element_id: str) -> dict:
...
+        Returns:
+            dict: same as `get_lane_coords` but overwrite xyz values for the lanes
+        """
+        lane_dict = self.get_lane_coords(element_id)
+        for lane_key in ["xyz_left", "xyz_right"]:
+            xyz_raw = lane_dict[lane_key]
+
+            # cumulative distance from the first point, including the first itself as 0
+            distances = np.cumsum(np.linalg.norm(np.diff(xyz_raw, axis=0), axis=-1))
+            distances = np.insert(distances, 0, 0)
```
anasrferreira, line 167: Wondering if "insert" in idx=0 is costly here, since it may imply moving all elements... tried to look up but couldn't get a good answer. Do you think they maybe a concatenation, or initializing distances with all zeros could be better? I may be way off though.

anasrferreira (approved):
Nice!

l5kit/l5kit/data/map_api.py
lucabergamini: I think (and I would like to stress the "think") that numpy stores doubled linked list here. Anyway, this function is LRU cached, so it should be fine :)

anasrferreira: That's great then.

l5kit/l5kit/data/map_api.py
```diff
@@ -134,6 +142,81 @@ def get_lane_coords(self, element_id: str) -> dict:
...
+        distances_left = np.insert(distances_left, 0, 0)
+
+        distances_right = np.cumsum(np.linalg.norm(np.diff(xyz_right, axis=0), axis=-1))
+        distances_right = np.insert(distances_right, 0, 0)
+
+        lane_dict["xyz_left"] = self.interpolate(xyz_left, distances_left, step, method)
+        lane_dict["xyz_right"] = self.interpolate(xyz_right, distances_right, step, method)
+
+        # compute midlane using mean of fixed length lanes
+        if method != InterpolationMethod.INTER_ENSURE_LEN:
```
anasrferreira, line 206: Perhaps add a comment to why the extra computation makes more sense if not using INTER_ENSURE_LEN. I almost forgot... and we had discussed Wed morning. So perhaps is useful to include it here. If you agree, but happy either way.

anasrferreira (approved):

anasrferreira (approved):

l5kit/l5kit/data/map_api.py
```diff
@@ -11,7 +12,14 @@
 ENCODING = "utf-8"
 
 
+class InterpolationMethod(IntEnum):
+    INTER_METER = 0  # fixed interpolation at a given step in meters
+    INTER_ENSURE_LEN = 1  # ensure we always get the same number of elements
+
+
 class MapAPI:
+    TR_FACES_COLORS = ["red", "green", "yellow"]
```
perone, line 21: I would either define the `InterpolationMethod` as a string list too, or change this `TR_FACES_COLORS` to an enum too (you can get the string value by the `.name()` method from the enum.

l5kit/l5kit/data/map_api.py
```diff
@@ -11,7 +12,14 @@
 ENCODING = "utf-8"
 
 
+class InterpolationMethod(IntEnum):
+    INTER_METER = 0  # fixed interpolation at a given step in meters
+    INTER_ENSURE_LEN = 1  # ensure we always get the same number of elements
```
perone, line 17: I think you'll need to add `#:` to autodoc be able to render this doc.

l5kit/l5kit/data/map_api.py
```diff
@@ -134,6 +144,92 @@ def get_lane_coords(self, element_id: str) -> dict:
...
+        INTER_METER: interpolate using step as a meter value over cumulative distances (variable len result)
+        INTER_ENSURE_LEN: interpolate using a variable step such that we always get step values
+        Args:
+            xyz (np.ndarray): XYZ coords
+            cum_dist (np.ndarray): distances from the first coords of xyz. Same length as xyz
+            step (float): param for the interpolation
+            method (InterpolationMethod): method to use to interpolate
+
+        Returns:
+            np.ndarray: the new interpolated coordinates
```
perone, line 160: I think we should start migrating the doc format to the same that avjadoo uses:
```
:param xyz: etc
:param cum_dist: etc
```
Just to have a single unified documentation format.

l5kit/l5kit/data/map_api.py
```diff
@@ -134,6 +144,92 @@ def get_lane_coords(self, element_id: str) -> dict:
...
+        if method == InterpolationMethod.INTER_ENSURE_LEN:
+            step = int(step)
+            assert step > 1, "step must be at least 2 with INTER_ENSURE_LEN"
+            steps = np.linspace(cum_dist[0], cum_dist[-1], step)
+
+        elif method == InterpolationMethod.INTER_METER:
+            assert step > 0, "step must be greater than 0 with INTER_FIXED"
+            steps = np.arange(cum_dist[0], cum_dist[-1], step)
+        else:
+            raise NotImplementedError("interpolation method unknown")
```
perone, line 171: I would complement the error message with something like:
```
interp_options = [m.name for m in InterpolationMethod]
raise NotImplementedError(f"Interpolation method unknown, options available: {interp_options}") 
```

That way it would throw the following error:
```
NotImplementedError: Interpolation method unknown, available: ['INTER_METER', 'INTER_ENSURE_LEN']
```

l5kit/l5kit/data/map_api.py
```diff
@@ -134,6 +144,92 @@ def get_lane_coords(self, element_id: str) -> dict:
...
+
+        xyz_inter = np.zeros((len(steps), 3), dtype=xyz.dtype)
```
perone, line 173: Consider changing the `np.zeros` to `np.empty` if you're replacing all values.

l5kit/l5kit/data/map_api.py
```diff
@@ -134,6 +144,92 @@ def get_lane_coords(self, element_id: str) -> dict:
...
+        xyz_inter[:, 0] = np.interp(steps, xp=cum_dist, fp=xyz[:, 0])
+        xyz_inter[:, 1] = np.interp(steps, xp=cum_dist, fp=xyz[:, 1])
+        xyz_inter[:, 2] = np.interp(steps, xp=cum_dist, fp=xyz[:, 2])
```
perone, line 176: I'm not sure about it but I think that the `np.interp` would allow you to do this on a single call, but have to double check.

l5kit/l5kit/data/map_api.py
```diff
@@ -134,6 +144,92 @@ def get_lane_coords(self, element_id: str) -> dict:
...
+        Perform an interpolation of the left and right lanes and compute the midlane.
+        See interpolate for details about the different interpolation methods
+
+        Args:
+            element_id (str): lane id
+            step (float): step param for the method
+            method (InterpolationMethod): one of the accepted methods
+
+        Returns:
+            dict: same as `get_lane_coords` but overwrite xyz values for the lanes
```
perone, line 196: Same thing on the doc formatting here.

l5kit/l5kit/rasterization/semantic_rasterizer.py
```diff
@@ -162,47 +116,55 @@ def render_semantic_map(
...
             for tl_id in lane_tl_ids.intersection(active_tl_ids):
-                if self.proto_API.is_traffic_face_colour(tl_id, "red"):
+                if self.mapAPI.is_traffic_face_colour(tl_id, "red"):
                     lane_type = "red"
-                elif self.proto_API.is_traffic_face_colour(tl_id, "green"):
+                elif self.mapAPI.is_traffic_face_colour(tl_id, "green"):
                     lane_type = "green"
-                elif self.proto_API.is_traffic_face_colour(tl_id, "yellow"):
+                elif self.mapAPI.is_traffic_face_colour(tl_id, "yellow"):
                     lane_type = "yellow"
```
perone, line 143: Would it be possible to have a `get_traffic_face_colour(tl_id)` ? So we don't have to check for all potential options.

l5kit/l5kit/rasterization/semantic_rasterizer.py
```diff
@@ -162,47 +116,55 @@ def render_semantic_map(
...
 
-        cv2.polylines(img, lanes_lines["default"], False, (255, 217, 82), lineType=cv2.LINE_AA, shift=CV2_SHIFT)
-        cv2.polylines(img, lanes_lines["green"], False, (0, 255, 0), lineType=cv2.LINE_AA, shift=CV2_SHIFT)
-        cv2.polylines(img, lanes_lines["yellow"], False, (255, 255, 0), lineType=cv2.LINE_AA, shift=CV2_SHIFT)
-        cv2.polylines(img, lanes_lines["red"], False, (255, 0, 0), lineType=cv2.LINE_AA, shift=CV2_SHIFT)
+            lanes_area = lanes_area.reshape((-1, INTERPOLATION_POINTS, 2))
+            cv2.polylines(img, lanes_area[lanes_mask["default"]], False, (255, 217, 82), **CV2_SUB_VALUES)
+            cv2.polylines(img, lanes_area[lanes_mask["red"]], False, (255, 0, 0), **CV2_SUB_VALUES)
+            cv2.polylines(img, lanes_area[lanes_mask["yellow"]], False, (255, 255, 0), **CV2_SUB_VALUES)
+            cv2.polylines(img, lanes_area[lanes_mask["green"]], False, (0, 255, 0), **CV2_SUB_VALUES)
```
perone, line 157: I would define a constant for the colours `(255, 217, 82), (255, 0, 0)`,  etc, associated w/ the `default`, `red`, etc.

l5kit/l5kit/data/map_api.py
```diff
@@ -200,6 +311,66 @@ def is_traffic_face_colour(self, element_id: str, colour: str) -> bool:
...
+                x_max = max(np.max(lane["xyz_left"][:, 0]), np.max(lane["xyz_right"][:, 0]))
+                y_max = max(np.max(lane["xyz_left"][:, 1]), np.max(lane["xyz_right"][:, 1]))
+
+                lanes_bounds = np.append(lanes_bounds, np.asarray([[[x_min, y_min], [x_max, y_max]]]), axis=0)
+                lanes_ids.append(element_id)
+
+            if self.is_crosswalk(element):
+                crosswalk = self.get_crosswalk_coords(element_id)
+                x_min = np.min(crosswalk["xyz"][:, 0])
+                y_min = np.min(crosswalk["xyz"][:, 1])
```
perone, line 360: I think you can reduce the calls here to have like `x_min, y_min = np.min(crosswalk["xyz"], axis=0)`

perone (approved):
Just minor comments 👍

l5kit/l5kit/data/map_api.py
```diff
@@ -200,6 +311,66 @@ def is_traffic_face_colour(self, element_id: str, colour: str) -> bool:
             return True
         return False
 
+    @lru_cache(maxsize=CACHE_SIZE)
+    def get_color_for_face(self, face_id: str) -> str:
```
yawei-ye, line 315:
```suggestion
    def get_traffic_light_state(self, tl_face_id: str) -> str:
```

l5kit/l5kit/data/map_api.py
```diff
@@ -176,6 +272,21 @@ def get_crosswalk_coords(self, element_id: str) -> dict:
 
         return {"xyz": xyz}
 
+    def is_traffic_light(self, element_id: str) -> bool:
```
yawei-ye, line 275:
```suggestion
    def has_traffic_light(self, element_id: str) -> bool:
```

l5kit/l5kit/data/map_api.py
```diff
@@ -176,6 +272,21 @@ def get_crosswalk_coords(self, element_id: str) -> dict:
...
+        """
+        Check if the element is a traffic light
```
yawei-ye, line 277:
```suggestion
        Check if the given map element has traffic light field. If yes, the map element is controlled by a traffic light.
```

l5kit/l5kit/data/map_api.py
```diff
@@ -176,6 +272,21 @@ def get_crosswalk_coords(self, element_id: str) -> dict:
...
+        Returns:
+            True if the element is a traffic light
+        """
+        element = self[element_id]
+        if not element.element.HasField("traffic_control_element"):
+            return False
+        traffic_el = element.element.traffic_control_element
+        return traffic_el.HasField("traffic_light") is True
+
     def is_traffic_face_colour(self, element_id: str, colour: str) -> bool:
```
yawei-ye, line 290:
```suggestion
    def check_traffic_light_state(self, element_id: str, tl_state: str) -> bool:
```

l5kit/l5kit/rasterization/semantic_rasterizer.py
```diff
@@ -162,47 +116,55 @@ def render_semantic_map(
...
                     lane_type = "yellow"
+            lanes_mask[lane_type][idx * 2 : idx * 2 + 2] = True
+
+        if len(lanes_area):
+            lanes_area = cv2_subpixel(transform_points(lanes_area.reshape((-1, 2)), raster_from_world))
 
-            lanes_lines[lane_type].extend([xy_left, xy_right])
+            for lane_area in lanes_area.reshape((-1, INTERPOLATION_POINTS * 2, 2)):
+                # need to for-loop otherwise some of them are empty
+                cv2.fillPoly(img, [lane_area], (17, 17, 31), **CV2_SUB_VALUES)
```
yawei-ye, line 151: where does (17, 17, 31) come from?

yawei-ye (commented):
Leave some suggestions on function names, otherwise LGTM

l5kit/l5kit/rasterization/semantic_rasterizer.py
yawei-ye: +1

l5kit/l5kit/rasterization/semantic_rasterizer.py
yawei-ye: +1

l5kit/l5kit/data/map_api.py
lucabergamini: so reST and not Google then...sigh I love the Google's one xD

l5kit/l5kit/rasterization/semantic_rasterizer.py
lucabergamini: no idea, it was there before I started working on this

l5kit/l5kit/data/map_api.py
lucabergamini: I will leave it as it is for now. I think we should aim for a single PR to change them all. I will start working on that

perone: 👍

l5kit/l5kit/data/map_api.py
lucabergamini: according to the docstring it's only 1D and it doesn't take axis as an argument :(

l5kit/l5kit/data/map_api.py
lucabergamini: all other (lanes, crosswalks) functions are `is` and not `has`. I think `is` makes more sense as you're checking what the element is in this case..

l5kit/l5kit/data/map_api.py
lucabergamini: no this tells you if the elements is a traffic light, there is no connection to a map element controlled by it

l5kit/l5kit/data/map_api.py
lucabergamini: state is something like `ACTIVE, INACTIVE, here we're just checking if the element is a traffic light face and which colour it has. It's a static info, not a dynamic one

l5kit/l5kit/data/map_api.py
yawei-ye: okay, I misunderstood it.

yawei-ye (commented):
Minor

yawei-ye (approved):
LGTM

perone (approved):
Looks good to me ! 👍

PR #202, opened 2020-11-23
Simplify filter_by_track_id function

lucabergamini:
Expose filter_by_track_id function and add tests for it

simplify implementation, better tests


yawei-ye (approved):
LGTM

PR #204, opened 2020-11-23
[Planning] Add model

lucabergamini:
Add model for drivenet.
Also, make rasteriser aware of their outputs, so that we can query them as:
```python
rasteriser = ...
channels = rasteriser.num_channels()
``` 
This will be used by the drivenet model


l5kit/l5kit/drivenet/model.py
```diff
@@ -0,0 +1,67 @@
...
+    def forward(self, data_batch: Dict[str, torch.Tensor]) -> Dict[str, torch.Tensor]:
+        image_batch = data_batch["image"]
+        outputs = self.model(image_batch)
+        batch_size = len(data_batch["image"])
+
+        if self.training:
+            if self.criterion is None:
+                raise NotImplementedError("Loss function is undefined.")
+
+            targets = (torch.cat((data_batch["target_positions"], data_batch["target_yaws"]), dim=2)).view(
```
yawei-ye, line 53: Can you add dimension for each tensor? Better readability
```
# [batch_size, num_timestamps, 2]

yawei-ye (approved):
LGTM!

PR #206, opened 2020-11-23
[DriveNet] Drivenet Open-loop Notebook

lucabergamini:


examples/drivenet/drivenet_open_loop.ipynb
```diff
@@ -0,0 +1,348 @@
...
+   "metadata": {},
+   "source": [
+    "# DriveNet Open Loop Evaluation\n",
+    "\n",
+    "**Note: this notebook assumes you've already run the [training notebook](./drivenet_train.ipynb) and stored your model successfully.**\n",
+    "\n",
+    "### What is open loop evaluation?\n",
+    "In open-loop evaluation we evaluate our model prediction as we follow the annotated ground truth.\n",
+    "\n",
+    "In each frame, we compare the predictions of our model against the ones annotated. This can be done with different metrics, such as ADE (Average Displacement Error) or FDE (Final Displacement Error). \n",
```
oliver-scheel, line 14: When not knowing these metrics here, I'd be compelled to google.
Remove the names here, maybe simply spell 'measure distance to ref. trajectory'?

examples/drivenet/drivenet_open_loop.ipynb
```diff
@@ -0,0 +1,348 @@
...
+    "\n"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "model_path = \"/var/folders/7f/mb3llzfs40b86yf2km8fml2h0000gp/T/drivenet.pt\"\n",
```
oliver-scheel, line 88: Leave this cryptic path?

examples/drivenet/drivenet_open_loop.ipynb
```diff
@@ -0,0 +1,348 @@
...
+   "source": [
+    "### Quantitative Evaluation: ADE, FDE and Angle Distance\n",
+    "\n",
+    "#### Positional Displacement\n",
+    "Average Displacement Error (ADE) and Final Displacement Error(FDE) are standard metrics used to evaluate future predictions for AVs.\n",
+    "\n",
+    "We can easily compute them by comparing predicted and annotated positions, which we have saved in the previous cell.\n",
+    "Additionally, we can plot histograms of their distributions across samples to better capture the variance of our error\n",
+    "\n",
+    "#### Angle Displacement\n",
```
oliver-scheel, line 167: Why is this useful? Bit less obvious to me than position - or rather, ADE and FDE are better motivated.

examples/drivenet/drivenet_open_loop.ipynb
```diff
@@ -0,0 +1,348 @@
...
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# Qualitative Evaluation: Visual Plots\n",
+    "\n",
+    "### Visualise Results\n",
+    "Clearly, manual inspection is still crucial to assess our model performance.\n",
```
oliver-scheel, line 218: Nit: sounds a bit like our metrics are really bad (which they are not)

examples/drivenet/drivenet_open_loop.ipynb
lucabergamini: it's for testing will be removed :)

examples/drivenet/drivenet_open_loop.ipynb
lucabergamini: my intuition is that if we don't include it we're not reporting any metrics on the yaw. In close loop later yaw is important for predictions, so it makes sense to check if it's reasonable imo

PR #207, opened 2020-11-24
[AVR-772] Add additional info for the competition ref system

lucabergamini:
Add additional info on the competition GT ref system

### DIFF
--- a/competition.md
+++ b/competition.md
@@ -51,11 +51,14 @@ Then, we rewrite this by substracting the maximum value x<sup>*</sup> from each 
 # Coordinates System for the competition
 Please refer to [this doc](https://github.com/lyft/l5kit/blob/master/coords_systems.md) for a full description of the different coordinate systems used in L5Kit.
 
-The ground truth coordinates for the competition are stored in the [world coordinate system](https://github.com/lyft/l5kit/blob/master/coords_systems.md#world-coordinate-system).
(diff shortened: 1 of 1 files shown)


competition.md
```diff
@@ -48,6 +48,15 @@ Then, we rewrite this by substracting the maximum value x<sup>*</sup> from each
 
 ![equation](https://latex.codecogs.com/gif.latex?%5Cbg_white%20%5Clarge%20LSE%28x_1%2C%20%5Cldots%2C%20x_n%29%20%3D%20x%5E*%20&plus;%20%5Clog%28e%5E%7Bx_1%20-%20x%5E%7B*%7D%7D%20&plus;%20%5Cldots%20&plus;%20e%5E%7Bx_n%20-%20x%5E%7B*%7D%7D%29)
 
+# Coordinates System for the competition
+Please refer to [this doc](https://github.com/lyft/l5kit/blob/master/coords_systems.md) for a full description of the different coordinate systems used in L5Kit.
+
+The ground truth coordinates for the competition are stored in the [world coordinate system](https://github.com/lyft/l5kit/blob/master/coords_systems.md#world-coordinate-system).
```
alcinos, line 54: Unless I'm missing something (based on my understanding of the example notebook), it seems that the ground-truth coordinates are actually displacements (computed from the last known position of the agent) in world coordinates. If this is true, this should be clearly stated here in my opinion, because the current formulation "ground truth coordinates" leads one to believe that absolute coordinates are expected.

Am I missing something?

lucabergamini: You're right, we should clearly state it :)

szagoruyko (approved):
looks good thanks Luca!

PR #209, opened 2020-11-27
Vectorized draw_boxes function

louis925:
* Vectorize draw_boxes function to speed up box_rasterizer
* Improve performance by about 10% to 40% on my local test

### DIFF
--- a/l5kit/l5kit/rasterization/box_rasterizer.py
+++ b/l5kit/l5kit/rasterization/box_rasterizer.py
@@ -7,7 +7,6 @@ from l5kit.data.zarr_dataset import AGENT_DTYPE
 
 from ..data.filter import filter_agents_by_labels, filter_agents_by_track_id
 from ..geometry import rotation33_as_yaw, transform_points
-from ..geometry.transform import yaw_as_rotation33
 from .rasterizer import EGO_EXTENT_HEIGHT, EGO_EXTENT_LENGTH, EGO_EXTENT_WIDTH, Rasterizer
 from .render_context import RenderContext
 from .semantic_rasterizer import CV2_SHIFT, cv2_subpixel
@@ -55,14 +54,14 @@ def draw_boxes(
     else:
         im = np.zeros((raster_size[1], raster_size[0], 3), dtype=np.uint8)
 
-    box_world_coords = np.zeros((len(agents), 4, 2))
-    corners_base_coords = np.asarray([[-1, -1], [-1, 1], [1, 1], [1, -1]])
+    corners_base_coords = (np.asarray([[-1, -1], [-1, 1], [1, 1], [1, -1]]) / 2)[None, :, :]
 
     # compute the corner in world-space (start in origin, rotate and then translate)
-    for idx, agent in enumerate(agents):
-        corners = corners_base_coords * agent["extent"][:2] / 2  # corners in zero
-        r_m = yaw_as_rotation33(agent["yaw"])
-        box_world_coords[idx] = transform_points(corners, r_m) + agent["centroid"][:2]
+    corners_m = corners_base_coords * agents["extent"][:, None, :2]  # corners in zero
+    s = np.sin(agents["yaw"])
+    c = np.cos(agents["yaw"])
+    rotation_m = np.moveaxis(np.array(((c, -s), (s, c))), 2, 0)
+    box_world_coords = np.einsum("bti,bji->btj", corners_m, rotation_m) + agents["centroid"][:, None, :2]
 
     box_raster_coords = transform_points(box_world_coords.reshape((-1, 2)), raster_from_world)
 
--- a/l5kit/l5kit/rasterization/box_rasterizer.py
+++ b/l5kit/l5kit/rasterization/box_rasterizer.py
@@ -54,12 +54,14 @@ def draw_boxes(
     else:
         im = np.zeros((raster_size[1], raster_size[0], 3), dtype=np.uint8)
 
-    corners_base_coords = (np.asarray([[-1, -1], [-1, 1], [1, 1], [1, -1]]) / 2)[None, :, :]
+    corners_base_coords = (np.asarray([[-1, -1], [-1, 1], [1, 1], [1, -1]]) * 0.5)[None, :, :]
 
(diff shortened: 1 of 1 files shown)


louis925:
Ok, looks like you guys like double quote...

louis925:
@lucabergamini the PR as promised

lucabergamini:
> Ok, looks like you guys like double quote...

it's [black](https://github.com/psf/black) :D

lucabergamini:
I'll take a look tomorrow morning (I'm at CET) :)

louis925:
No worry it is holiday now 🙂

l5kit/l5kit/rasterization/box_rasterizer.py
```diff
@@ -55,14 +54,14 @@ def draw_boxes(
...
 
     # compute the corner in world-space (start in origin, rotate and then translate)
-    for idx, agent in enumerate(agents):
-        corners = corners_base_coords * agent["extent"][:2] / 2  # corners in zero
-        r_m = yaw_as_rotation33(agent["yaw"])
-        box_world_coords[idx] = transform_points(corners, r_m) + agent["centroid"][:2]
+    corners_m = corners_base_coords * agents["extent"][:, None, :2]  # corners in zero
+    s = np.sin(agents["yaw"])
+    c = np.cos(agents["yaw"])
+    rotation_m = np.moveaxis(np.array(((c, -s), (s, c))), 2, 0)
```
lucabergamini, line 63: this part is also equivalent (tested) to:
```python
 rotation_m = np.moveaxis(np.array(((c, s), (-s, c))), 2, 0)
 box_world_coords = corners_m @ rotation_m + agents["centroid"][:, None, :2]
```
Which one do you think is the most clear in your opinion?

I'm not familiar with einsum so for me that part is magic, but at the same time my solution has a clockwise matrix to avoid transposing the points. I don't really know lol

l5kit/l5kit/rasterization/box_rasterizer.py
```diff
@@ -55,14 +54,14 @@ def draw_boxes(
     else:
         im = np.zeros((raster_size[1], raster_size[0], 3), dtype=np.uint8)
 
-    box_world_coords = np.zeros((len(agents), 4, 2))
-    corners_base_coords = np.asarray([[-1, -1], [-1, 1], [1, 1], [1, -1]])
+    corners_base_coords = (np.asarray([[-1, -1], [-1, 1], [1, 1], [1, -1]]) / 2)[None, :, :]
```
lucabergamini, line 57: any reasons to move the `\2`? I'm fine in having it here, just asking :)

lucabergamini (commented):
minor, but amazing work :)!

l5kit/l5kit/rasterization/box_rasterizer.py
louis925: I think I was trying to reduce the number of operations in each steps. So `corners_base_coords` was outside the `draw_boxes`. But I found it actually don't contribute too much so I move them back into the function. We can put the `/ 2` anywhere if you think that is more readable.

l5kit/l5kit/rasterization/box_rasterizer.py
louis925: Good idea. I think matrix multiplication is still faster than einsum because there might be some optimization for it. I will change that.

louis925: Speed is similar to the einsum. So I changed to using matrix multiplication for easier read.

l5kit/l5kit/rasterization/box_rasterizer.py
lucabergamini: I'd move it back to line 60 to make it clear we're using half of the extent. We can also replace it with `*0.5` (mul is generally faster but it won't probably matter here as we don't have thousands of corners)

l5kit/l5kit/rasterization/box_rasterizer.py
```diff
@@ -55,14 +54,14 @@ def draw_boxes(
...
 
     # compute the corner in world-space (start in origin, rotate and then translate)
-    for idx, agent in enumerate(agents):
-        corners = corners_base_coords * agent["extent"][:2] / 2  # corners in zero
-        r_m = yaw_as_rotation33(agent["yaw"])
-        box_world_coords[idx] = transform_points(corners, r_m) + agent["centroid"][:2]
+    corners_m = corners_base_coords * agents["extent"][:, None, :2]  # corners in zero
+    s = np.sin(agents["yaw"])
+    c = np.cos(agents["yaw"])
+    rotation_m = np.moveaxis(np.array(((c, s), (-s, c))), 2, 0)
```
lucabergamini, line 63: can we add a comment here saying that this is clockwise because it's right-multiplied and not left-multiplied later (and therefore we're still rotating conter-clockwise)?

lucabergamini (dismissed):
LGTM! feel free to address the comments or leave the code as it is. I'll merge it on Monday :) 

Amazing work, thanks a lot!

l5kit/l5kit/rasterization/box_rasterizer.py
louis925: My understanding is that the corner base should be at +-0.5. So later when we multiply by extend, it will be at +-0.5*extend. But I guess they are equivalent.

louis925: In a special case, you could also place your corners at -0.3*extent and 0.7*extent. In this case multiply extend by 0.5 or 0.3 will not work.

lucabergamini: > My understanding is that the corner base should be at +-0.5. So later when we multiply by extend, it will be at +-0.5*extend. But I guess they are equivalent.

They are equivalent. What you want to achieve is place the corners where they should be before rotating and translating



> Also, I think it is still faster to multiply 0.5 to corners_base_coords before it is exploded by the agent extent since after the multiplication to the `agent['extent']`, the matrix becomes (N agents) * (4 corners).

Yes, it should be faster. let leave it there then

lucabergamini (approved):

PR #210, opened 2020-11-27
[AVR-762] refactor agent_sampling

lucabergamini:
Refactor `generate_agent_sample` by breaking it into multiple functions

TODO:
- [x] `get_agent_context` tests
- [x] see if we can factor something else out


l5kit/l5kit/sampling/agent_sampling.py
```diff
@@ -15,6 +15,94 @@
...
+    tl_faces = tl_faces[tl_slice].copy()
+    # sync interval with the traffic light faces array
+    history_frames["traffic_light_faces_index_interval"] -= tl_slice.start
+    history_tl_faces = filter_tl_faces_by_frames(history_frames, tl_faces)
+    future_tl_faces = filter_tl_faces_by_frames(future_frames, tl_faces)
+
+    return history_frames, future_frames, history_agents, future_agents, history_tl_faces, future_tl_faces
+
+
+def compute_agent_speed(
```
yawei-ye, line 74:
```suggestion
def compute_agent_velocity(
```
We are computing velocity (v_x, v_y) and not speed (scalar)

l5kit/l5kit/sampling/agent_sampling.py
```diff
@@ -15,6 +15,94 @@
...
+    history_positions_m: np.ndarray, future_positions_m: np.ndarray, history_step_time: float, future_step_time: float
+) -> Tuple[np.ndarray, np.ndarray]:
+    """
+    compute estimated velocities by finite differentiation on future positions
```
yawei-ye, line 78:
```suggestion
    Compute estimated velocities by finite differentiation on future positions (pos(T+t) - pos(T))/t. This simple approach gives less than 0.5% velocity difference compared to (pos(T+t) - pos(T-t))/2t on v1.1/sample.zarr.tar.
```

yawei-ye (commented):
Some minor comments, otherwise looking good!

l5kit/l5kit/sampling/agent_sampling.py
```diff
@@ -15,6 +15,94 @@
...
+    tl_faces = tl_faces[tl_slice].copy()
+    # sync interval with the traffic light faces array
+    history_frames["traffic_light_faces_index_interval"] -= tl_slice.start
+    history_tl_faces = filter_tl_faces_by_frames(history_frames, tl_faces)
+    future_tl_faces = filter_tl_faces_by_frames(future_frames, tl_faces)
+
+    return history_frames, future_frames, history_agents, future_agents, history_tl_faces, future_tl_faces
+
+
+def compute_agent_velocity(
```
anasrferreira, line 74: Seems like this should be a more generic function. Instead of computing both history and future, a function that just gets an array of positions + delta time and then computes the finite differences (this means it would be called twice, once for history, once for future).

l5kit/l5kit/sampling/agent_sampling.py
```diff
@@ -15,6 +15,94 @@
...
+    Args:
+        history_positions_m (np.ndarray): history XY positions in meters
+        future_positions_m (np.ndarray): future XY positions in meters
+        history_step_time (np.ndarray): length of a step in second
+        future_step_time (np.ndarray): length of a step in second
+
+    Returns:
+        Tuple[np.ndarray, np.ndarray]: history and future XY speeds
+
+    """
```
anasrferreira, line 91: consider assert step_time_s > kEpsilon

l5kit/l5kit/sampling/agent_sampling.py
```diff
@@ -47,48 +135,44 @@ def generate_agent_sample(
         selected_track_id (Optional[int]): Either None for AV, or the ID of an agent that you want to
         predict the future of. This agent is centered in the raster and the returned targets are derived from
         their future states.
-        raster_size (Tuple[int, int]): Desired output raster dimensions
-        pixel_size (np.ndarray): Size of one pixel in the real world
-        ego_center (np.ndarray): Where in the raster to draw the ego, [0.5,0.5] would be the center
+        render_context (RenderContext): The context for rasterisation
```
anasrferreira, line 138: I actually wonder if render context should be removed, or Optional, since rasterizer can be optional. In any case seems like this is not in scope for this PR.

l5kit/l5kit/sampling/agent_sampling.py
```diff
@@ -15,6 +15,94 @@
...
+
+    # current position is included in history positions
+    # [history_num_frames, 2]
+    history_positions_diff_m = np.diff(history_positions_m, axis=0)
+    # [history_num_frames, 2]
+    history_vels_mps = np.float32(history_positions_diff_m / history_step_time)
+
+    return history_vels_mps, future_vels_mps
+
+
```
anasrferreira, line 105: Seems like the history_step_time here may be misused, right? Example, the user needs to know that history_step_size and history_step_time must be consistent with each other, otherwise the returned frames may represent different things. Or am I missing something?
Also, this is probably not something in scope for this PR either, but the way history_step_time and history_step_size are used seems like they are related with the dataset's frame rate.

anasrferreira (approved):

l5kit/l5kit/sampling/agent_sampling.py
lucabergamini: Probably yes (it's also inside the rasteriser anyway)

l5kit/l5kit/sampling/agent_sampling.py
lucabergamini: I see your point here. We don't have for now another case where we compute finite difference and divide them by something. I'm leaning towards leaving this as it is for now and change it in the future if we find a common use case

l5kit/l5kit/sampling/agent_sampling.py
lucabergamini: yes this is a very good point. `history_step_size` and `future_step_time` are not properly check for values greater than 1 in the codebase. I'm actually tempted to remove them as they can be misused

lucabergamini: #213 will remove that para

l5kit/l5kit/tests/sampling/displace_test.py
```diff
@@ -16,7 +16,7 @@ def base_displacement(zarr_dataset: ChunkedDataset, cfg: dict) -> np.ndarray:
         ref_frame["ego_translation"][:2], rotation33_as_yaw((ref_frame["ego_rotation"]))
     )
 
-    future_coords_offset, *_ = _create_targets_for_deep_prediction(
+    future_coords_offset, *_ = get_relative_poses(
```
yawei-ye, line 19:
```suggestion
    future_positions, *_ = get_relative_poses(
```

yawei-ye (approved):
Minor, otherwise LGTM.
Like the new function names!

yawei-ye (approved):

PR #212, opened 2020-12-03
[AVR-763] Perform flip in the render_context

lucabergamini:
Perform flip of y axis in the `render_contex` (while creating the matrix)
Things to do:
- [x] add config flag to preserve previous models
- [x] read the config value and pass it to the render_context
- [x] fix tests ~(should be just the one with the render_context itself)~ spoiler: it was not
- [x] fix notebooks
- [x] train new models
- [x] test new models
- [x] upload new models

I've removed a comment in the satellite, maybe I should replace it with something else

PR #213, opened 2020-12-07
[AVR-786] Remove step-size from codebase

lucabergamini:
Remove step-size (both past and future) from codebase

Steps greater than 1 were not supported anyway and were really
bug-prone.

The same functionality can be achieved by subsampling the result:

- the same amount of data is loaded in the two cases (zarr still load
  the entire chunk clearly even for strided slices);
- the required pre-processing in greater, but that's not a bottleneck
  compared with data loading and rasterisation


examples/agent_motion_prediction/agent_motion_config.yaml
```diff
@@ -7,12 +7,8 @@ model_params:
   model_architecture: "resnet50"
 
   history_num_frames: 0
-  history_step_size: 1
-  history_delta_time: 0.1
-
   future_num_frames: 50
-  future_step_size: 1
-  future_delta_time: 0.1
+  delta_time: 0.1
```
yawei-ye, line 11:
```suggestion
  step_time: 0.1
```
How about call it `step_time`? Be consistent with the variable name used in the code.

l5kit/l5kit/dataset/ego.py
```diff
@@ -53,11 +53,8 @@ def __init__(
             generate_agent_sample,
             render_context=render_context,
             history_num_frames=cfg["model_params"]["history_num_frames"],
-            history_step_size=cfg["model_params"]["history_step_size"],
-            history_step_time=cfg["model_params"]["history_delta_time"] * cfg["model_params"]["history_step_size"],
             future_num_frames=cfg["model_params"]["future_num_frames"],
-            future_step_size=cfg["model_params"]["future_step_size"],
-            future_step_time=cfg["model_params"]["future_delta_time"] * cfg["model_params"]["future_step_size"],
+            step_time=cfg["model_params"]["delta_time"],
```
yawei-ye, line 57:
```suggestion
            step_time=cfg["model_params"]["step_time"],
```

yawei-ye (commented):
Great, it is so much cleaner! After this PR being merged, can you update AVR side?

lucabergamini:
> After this PR being merged, can you update AVR side?

should we publish a new release maybe with changes from this and #212 ?

yawei-ye (approved):
LGTM!

yawei-ye:
> > After this PR being merged, can you update AVR side?
> 
> should we publish a new release maybe with changes from this and #212 ?

Yes, we should. Also bump a major version in L5kit since it is a breaking change

PR #214, opened 2020-12-09
[AVR-789] add function to cast str to byte list

lucabergamini:
add function to cast str to byte list and apply it on `host_id`

TODO:
- [x] add test for function
- [x] fix notebooks


yawei-ye (approved):
LGTM

PR #219, opened 2020-12-21
[AVR-837] apply batch transform to sampling

lucabergamini:
Apply batch transform to positions. This saves N-1 transforms call for N future/history frames.

Add a note about the return dtype, as it gets casted from `float64` to float32` inside.
Question: should we cast here or later (provided we need to do it for torch)?

### DIFF
--- a/l5kit/l5kit/sampling/agent_sampling.py
+++ b/l5kit/l5kit/sampling/agent_sampling.py
@@ -9,7 +9,7 @@ from ..data import (
     get_tl_faces_slice_from_frames,
 )
 from ..data.filter import filter_agents_by_frames, filter_agents_by_track_id
-from ..geometry import angular_distance, compute_agent_pose, rotation33_as_yaw, transform_point
+from ..geometry import angular_distance, compute_agent_pose, rotation33_as_yaw, transform_points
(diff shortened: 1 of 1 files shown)


l5kit/l5kit/sampling/agent_sampling.py
```diff
@@ -148,11 +151,15 @@ def get_relative_poses(
...
-        yaws_rad[i] = angular_distance(agent_yaw_rad, current_agent_yaw)
+        positions_m[i] = agent_centroid_m
+        yaws_rad[i] = agent_yaw_rad
         extents_m[i] = agent_extent
         availabilities[i] = 1.0
-    return positions_m, yaws_rad, extents_m, availabilities
+
+    # batch transform to speed up
+    positions_m = transform_points(positions_m, agent_from_world) * availabilities[:, None]
+    yaws_rad = angular_distance(yaws_rad, current_agent_yaw) * availabilities[:, None]
```
szagoruyko, line 161: `np.newaxis` is more readable than `None`

szagoruyko (approved):
LGTM, one minor comment

szagoruyko (approved):

PR #221, opened 2021-01-06
[Fix] Satellite Image Transform for Non-Squared Raster

pascal-pfeiffer:
Non squared raster settings cause a wrong transform in the satellite images.
This patch should fix the issue. 

See the images below.

Before patch:
old_loader

After patch:
new_loader


lucabergamini (approved):
LGTM! I'll test it quickly tomorrow and merge it if nothing comes up! Amazing work @pascal-pfeiffer

PR #222, opened 2021-01-06
Possible Speedup: Move Normalization to GPU

pascal-pfeiffer:
Image normalization (cast to float and divide by 255) on GPU for a ~2x speedup when using larger history counts

Per sample timings on a mediocre machine (448x224 raster, 30 history frames):
77.9 ms ± 2.09 ms per loop (mean ± std. dev. of 7 runs, 10 loops each) with normalization (/255) on CPU
49.9 ms ± 2.61 ms per loop (mean ± std. dev. of 7 runs, 10 loops each) without normalization on CPU (must be done on GPU)


lucabergamini:
Although the speedup is massive this is quite a breaking change for the codebase:
- old models won't work after this unless you patch all the lines where data is forwarded to them;
- it's very easy to forget to apply this normalisation as there is no general rule about where to do it (if it's not done at data level).

pascal-pfeiffer:
I could think of a newly introduced config flag for gpu_normalization. 
- If true, the normalization would be done on the GPU and the user would benefit from the speedup. 
- If false (default value) nothing would change and the normalization would be done on the CPU

PR #226, opened 2021-01-15
Integration of the collision detection metric

perone:
This PR will:
- Add the functional `detect_collision` from the closed-loop evaluator;
- Adding other dependencies of the `detect_collision`: `_ego_agent_within_range`, `_get_bounding_box`, `_get_sides`;
- Add unit-testing for each dependency and the `detect_collision` as well;


perone:
meow @lucabergamini 
meow @yawei-ye

l5kit/l5kit/evaluation/metrics.py
```diff
@@ -1,6 +1,8 @@
-from typing import Callable
```
lucabergamini, line 1: are we sure we want to pile all metrics here? should we maybe create a new dedicated package?

l5kit/l5kit/evaluation/metrics.py
```diff
@@ -303,3 +305,131 @@ def final_displacement_error_mean(
...
+    :return: array with True if within range, False otherwise (shape: N)
+    """
+    distance = np.linalg.norm(ego_centroid - agent_centroid, axis=-1)
+    norm_ego = np.linalg.norm(ego_extent[..., :2], axis=-1)
+    norm_agent = np.linalg.norm(agent_extent[..., :2], axis=-1)
+    max_range = 0.5 * (norm_ego + norm_agent)
+    return distance < max_range
+
+
+def _get_bounding_box(centroid: np.ndarray, yaw: np.ndarray,
```
lucabergamini, line 338: `/planning/utils.py` already has these functions as we're using collisions in a planning notebook. Should we remove them and fix the notebook to use these?

I can also do it on top of this PR

l5kit/l5kit/evaluation/metrics.py
```diff
@@ -303,3 +305,131 @@ def final_displacement_error_mean(
...
+# be moved to an abstraction of the Agent later. Such as in the example below:
+#
+#  ego = Agent(...)
+#  agent = Agent(...)
+#  bbox = ego.get_boundingbox()
+#  within_range = ego.within_range(agent)
+#  sides = ego.get_sides()
+
+
+def _ego_agent_within_range(ego_centroid: np.ndarray, ego_extent: np.ndarray,
```
lucabergamini, line 320: the original function was vectorised over agents, this one is not. Is the idea to refactor it later when we design the metric framework?

l5kit/l5kit/evaluation/metrics.py
```diff
@@ -303,3 +305,131 @@ def final_displacement_error_mean(
...
+    max_range = 0.5 * (norm_ego + norm_agent)
+    return distance < max_range
+
+
+def _get_bounding_box(centroid: np.ndarray, yaw: np.ndarray,
+                      extent: np.ndarray,) -> Polygon:
+    """This function will get a shapely Polygon representing the bounding box
+    with an optional buffer around it.
+
+    :param centroid: centroid of the agent
```
lucabergamini, line 343: shapes missing

lucabergamini (approved):
minors. This will change later to conform to the new framework for metrics right?

l5kit/l5kit/evaluation/metrics.py
perone: My plan was to have all metrics here unless there is a significant semantic difference among them, but I haven't checked each one of them. Do you think we should split them into different modules ? For me it is ok, but if we can have all metrics together it would be simple to use and find them.

l5kit/l5kit/evaluation/metrics.py
perone: Wasn't aware they were on l5kit as well, I'll check if they have testing/docs there and move these into the `utils.py` in case they don't. I think it would be nice to have a module separate only for geometrical operations like these, also because they aren't specific for planning only.

l5kit/l5kit/evaluation/metrics.py
perone: This one is vectorized, right ? The one in `avresearch` is not.

l5kit/l5kit/evaluation/metrics.py
perone: Will fix

l5kit/l5kit/evaluation/metrics.py
lucabergamini: ah yes sorry. The one in avresearch is not, however the one I ported is vectorised (but in a different way).

A far as I can see from the signature this function works with N ego and N agents, is that correct? In general we can't assume that each ego will have N agents though (the number is variable)

perone: It can work with N ego and M agents, sorry the doc wasn't very clear, I'll fix it and use it on the `detect_collision` and submit another PR, should be clear then.

lucabergamini: In general, you will have an array of N ego and N list of agents where each one has a different length

perone:
@lucabergamini take one more look, I reverted to the same behavior of the within range function, except that it returns a mask instead of filtering the agents. I re-organized where the functions are and fixed the notebook too.

l5kit/l5kit/evaluation/metrics.py
```diff
@@ -1,7 +1,10 @@
-from typing import Callable
+from enum import IntEnum
+from typing import Callable, Optional, Tuple
 
 import numpy as np
 
+from l5kit.planning import utils
```
lucabergamini, line 6: at this point I would just move the stuff we need here and delete `l5kit.planning.utils`. My suggestion was more about not replicating code (and fix the notebook) than to keep that file. Sorry I should have clarified that from the start :)

l5kit/l5kit/planning/utils.py
```diff
@@ -26,41 +54,19 @@ def _get_sides(bbox: Polygon) -> Tuple[LineString, LineString, LineString, LineS
...
-    Computes whether a collision occurred between ego and any another agent.
-    Also computes the type of collision: rear, front, or side.
-    For this, we compute the intersection of ego's four sides with a target
-    agent and measure the length of this intersection. A collision
-    is classified into a class, if the corresponding length is maximal,
-    i.e. a front collision exhibits the longest intersection with
-    egos front edge.
+    :param ego_centroid: the ego centroid (shape: N, 2)
+    :param ego_extent: the ego extent (shape: N, 3)
+    :param agent_centroid: the agent centroid (shape: M, 2)
```
lucabergamini, line 64: can this work when N!=M?  locally the line
```python
np.linalg.norm(ego_centroid - agent_centroid, axis=-1)
```
crashes when they are different.

Can you add a test for that in `test_within_range`?

lucabergamini (commented):
minors, sorry again for the confusion

l5kit/l5kit/evaluation/metrics.py
perone: No problem, let's merge this one, and then later I can move things around, there are things we should probably move to a "geometry" package or something similar, but we can discuss that later.

l5kit/l5kit/planning/utils.py
perone: Don't know if understood it correctly, but this function should be the same as the original, do you want me to change the behavior of it ?

lucabergamini (approved):

PR #229, opened 2021-01-20
Fixing the issue with the norm over heights

perone:
Fix for the within range function used by the collision detection metric (norm was being calculated over heights of agents).


lucabergamini (approved):
LGTM! thanks for fixing this!

PR #234, opened 2021-02-03
Return `recent_speed` when possible.

blazejosinski:
Context: the `speed` parameter contained information about the future, thus allowing them to leak to the model. I substitute it with the history speed, which can be computed when history is given.

### DIFF
--- a/l5kit/l5kit/sampling/agent_sampling.py
+++ b/l5kit/l5kit/sampling/agent_sampling.py
@@ -258,7 +258,7 @@ def generate_agent_sample(
 
     history_vels_mps, future_vels_mps = compute_agent_velocity(history_positions_m, future_positions_m, step_time)
 
-    return {
+    result = {
         "frame_index": state_index,
         "image": input_im,
         "target_positions": future_positions_m,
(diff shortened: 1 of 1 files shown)


l5kit/l5kit/sampling/agent_sampling.py
```diff
@@ -276,8 +276,10 @@ def generate_agent_sample(
         "world_from_agent": world_from_agent,
         "centroid": agent_centroid_m,
         "yaw": agent_yaw_rad,
-        "speed": np.linalg.norm(future_vels_mps[0]),
         "extent": agent_extent_m,
         "history_extents": history_extents,
         "future_extents": future_extents,
     }
+    if len(history_vels_mps) > 0:
```
lucabergamini, line 283: does `history` in the name make sense given it's only one step? Is it because we want to emphasise that it is computed from past?

lucabergamini (approved):
LGTM, 1 nit

l5kit/l5kit/sampling/agent_sampling.py
blazejosinski: Correct. "past_speed" probably sounds better, will fix.

blazejosinski: "recent_speed" sounds even better, IMHO.

blazejosinski:
PTAL @lucabergamini

lucabergamini (approved):

PR #239, opened 2021-02-21
Adding Google colab badges into the README.md

perone:
This PR will add the Google colab badges into the README.md


lucabergamini (approved):
LGTM, but because our notebooks don't download the required data this doesn't really work right now. Should we address that?

perone:
Yes, this is just to prepare for later. We need to actually deal now with the license issue, I think one alternative would be to add a button on the notebook with acceptance of it or something like that. I will double-check that, if you already have the links on hand that are required for download please paste here or add in the top of the notebooks, so then later I can add the downloading and license acceptance part. Thanks !

PR #240, opened 2021-02-25
Add speed perturbation

yawei-ye:
# Changes
- Add speed perturbation in ackerman perturbation

### DIFF
--- a/l5kit/l5kit/kinematic/ackerman_perturbation.py
+++ b/l5kit/l5kit/kinematic/ackerman_perturbation.py
@@ -56,17 +56,6 @@ class AckermanPerturbation(Perturbation):
             )
 
     def perturb(self, history_frames: np.ndarray, future_frames: np.ndarray) -> Tuple[np.ndarray, np.ndarray]:
-        if np.random.rand() >= self.perturb_prob:
-            return history_frames.copy(), future_frames.copy(), (1.0, 0.0)
-
-        (
-            lateral_offset_m,
-            longitudinal_offset_m,
-            yaw_offset_rad,
-            speed_multiplier,
-            speed_offset,
-        ) = self.random_offset_generator()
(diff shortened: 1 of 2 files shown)


l5kit/l5kit/kinematic/ackerman_perturbation.py
```diff
@@ -93,7 +102,7 @@ def perturb(self, history_frames: np.ndarray, future_frames: np.ndarray) -> Tupl
         x0 = trajectory[curr_frame_idx, 0] + position_offset_m[0]
         y0 = trajectory[curr_frame_idx, 1] + position_offset_m[1]
         r0 = trajectory[curr_frame_idx, 2] + yaw_offset_rad
-        v0 = gv[0]
+        v0 = curr_displacement * speed_multiplier + speed_offset
```
blazejosinski, line 105: I think that speed_offset should be in meters per second. Here we should divide it by step_time.

l5kit/l5kit/kinematic/ackerman_perturbation.py
```diff
@@ -76,6 +68,23 @@ def perturb(self, history_frames: np.ndarray, future_frames: np.ndarray) -> Tupl
         curr_frame_idx = num_history_frames - 1
         displacements = np.linalg.norm(np.diff(trajectory[curr_frame_idx:, :2], axis=0), axis=1)
 
+        if num_history_frames > 1:
```
blazejosinski, line 71: IMHO we should explain this is the function docs, that it has different behavior, depending on availability of history frames.

blazejosinski (approved):

l5kit/l5kit/kinematic/ackerman_perturbation.py
yawei-ye: will make the change

perone:
Can I close this @yawei-ye ?

PR #248, opened 2021-03-18
Simulation Utils

lucabergamini:
branch for simulations utils:

* Path prior support: support to rasterise a path prior as a new raster channel

### DIFF
--- a/l5kit/l5kit/dataset/ego.py
+++ b/l5kit/l5kit/dataset/ego.py
@@ -21,6 +21,7 @@ class EgoDataset(Dataset):
             zarr_dataset: ChunkedDataset,
             rasterizer: Rasterizer,
             perturbation: Optional[Perturbation] = None,
+            render_path_prior: bool = False,
     ):
         """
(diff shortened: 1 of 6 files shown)


l5kit/l5kit/dataset/ego.py
```diff
@@ -21,6 +21,7 @@ def __init__(
             zarr_dataset: ChunkedDataset,
             rasterizer: Rasterizer,
             perturbation: Optional[Perturbation] = None,
+            render_path_prior: bool = False,
```
yawei-ye, line 24: Do we need to add render_path_prior at Line 168?: https://github.com/lyft/l5kit/pull/248/files#diff-e40a85c9573b89ac8b28c54ddf84835cad03fdcc9760a35762cd95ffff7d464bL168

lucabergamini: ah yeah that's a good point! I really don't like passing this arg...another option would be to add it in the `cfg`

PR #249, opened 2021-03-18
add signed speed

yawei-ye:
# Changes
- add signed speed
- [x] add unit test

### DIFF
--- a/l5kit/l5kit/sampling/agent_sampling.py
+++ b/l5kit/l5kit/sampling/agent_sampling.py
@@ -99,9 +99,14 @@ def compute_agent_velocity(
     return history_vels_mps, future_vels_mps
 
 
-def compute_signed_speeds(velocities_mps: np.ndarray) -> np.ndarray:
-    # compute signed speed, direction is based on x
-    signed_speeds_mps = np.sign(velocities_mps[:, 0]) * np.linalg.norm(velocities_mps, axis=1)
+def compute_signed_speeds(vels_mps: np.ndarray) -> np.ndarray:
+    """ Compute signed speeds by computing the 2-norm of the given velocities, the sign / direction of the speed is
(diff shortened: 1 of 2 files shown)


l5kit/l5kit/sampling/agent_sampling.py
```diff
@@ -99,6 +99,12 @@ def compute_agent_velocity(
     return history_vels_mps, future_vels_mps
 
 
+def compute_signed_speeds(velocities_mps: np.ndarray) -> np.ndarray:
```
lucabergamini, line 102: docstring for this

l5kit/l5kit/sampling/agent_sampling.py
```diff
@@ -275,6 +282,8 @@ def generate_agent_sample(
         "target_positions": future_positions_m,
         "target_yaws": future_yaws_rad,
         "target_velocities": future_vels_mps,
+        "target_speeds": future_speeds_mps,
```
lucabergamini, line 285: so this will be the new target for speed right?

lucabergamini (commented):
LGTM!, one minor

l5kit/l5kit/sampling/agent_sampling.py
yawei-ye: correct

yawei-ye: by the way, it is of shape [num_future_frames] and can be used to train model with longer time horizon.
I created another `"target_speed": future_speeds_mps[:1]` just for not breaking our existing model trained with single speed

lucabergamini: yeah I noticed that, good point

lucabergamini (approved):

PR #251, opened 2021-03-30
[AVR-969] Add bokeh visualiser

lucabergamini:
Add a visualiser based on html bokeh.

TODO:

- [ ] get to a working version
- [ ] add test
- [ ] add TLs signal
- [ ] add different colours for agents (based on type)
- [ ] add toggable trajectories
- [ ] add gif exporter (somehow)
- [ ] add mp4 exporter (somehow)
- [ ] discuss support for multiple trajectories per agent


l5kit/l5kit/visualization/visualiser/visualiser.py
```diff
@@ -0,0 +1,133 @@
...
+        ],
+    )
+    out: List[Dict[str, ColumnDataSource]] = []
+
+    trajectories_labels = np.unique([traj.legend_label for frame in frames for traj in frame.trajectories])
+
+    for frame_idx in range(len(frames)):
+        ego = frames[frame_idx].ego
+
+        ego_dict = {field_name: [field] for field, field_name in zip(ego, EgoVisualisation._fields)}
```
perone, line 49:
```suggestion
        ego_dict = {k:[v] for k, v in ego._asdict().items()}
```

l5kit/l5kit/visualization/visualiser/visualiser.py
```diff
@@ -0,0 +1,133 @@
...
+
+        agents_dict = {k: [] for k in AgentVisualisation._fields}
```
perone, line 51:
```suggestion
        agents_dict = defaultdict(list)
```

l5kit/l5kit/visualization/visualiser/common.py
```diff
@@ -0,0 +1,48 @@
...
+    ys: np.ndarray
+    color: str
+
+
+class AgentVisualisation(NamedTuple):
+    xs: np.ndarray
+    ys: np.ndarray
+    color: str
+    track_id: int
+    type: str
```
perone, line 23: `type()` is already a reserved function, I would consider changing the name here.

l5kit/l5kit/visualization/visualiser/common.py
```diff
@@ -0,0 +1,48 @@
...
+
+class EgoVisualisation(NamedTuple):
+    xs: np.ndarray
+    ys: np.ndarray
+    color: str
+    center_x: float
+    center_y: float
+
+
+class TrajectoryVisualisation(NamedTuple):
```
perone, line 35: I would stick with `Visualization` instead of `Visualisation`.

l5kit/l5kit/visualization/visualiser/visualiser.py
```diff
@@ -0,0 +1,133 @@
...
+        for agent in frames[frame_idx].agents:
+            for k, v in zip(AgentVisualisation._fields, agent):
```
perone, line 53:
```suggestion
            for k, v in agent._asdict().items():
```
Same for the others below. I would consider adding a function to do that since we're repeating the code below.

l5kit/l5kit/visualization/visualiser/visualiser.py
```diff
@@ -0,0 +1,133 @@
...
+    slider.js_on_change("value", slider_callback)
+
+    f.legend.location = "top_left"
+    f.legend.click_policy = "hide"
+
+    layout = column(f, slider)
+    if save_path is not None:
+        save(layout, filename=save_path, title=f"{scene_index}", resources=Resources(resources))
+    if do_show:
+        show(layout)
```
perone, line 133: I would remove this from here for the same reason as the `output_notebook()`, it should be the responsibility of the caller to decide what to do. We can just return the `layout` here and also remove the `resources` parameter from the function (that is also only being used here).

perone (commented):
Thanks, just added some comments regarding the structure of the visualization and some minors.

l5kit/l5kit/visualization/visualiser/visualiser.py
lucabergamini: function added

perone (approved):
We can move the visualizer to a class later, but let's merge it now to unblock, and then later we can think better about it. Thanks !

PR #253, opened 2021-03-31
[AVR-970] Add Simulation classes and Loop

lucabergamini:
Add objects required for unroll simulation (ego and agents)

- [x] add main class
- [x] add test for ego get
- [x] add test for ego set
- [x] add test for agents get
- [x] add test for agents set
- [x] add simulation loop
- [x] add tests for simulation loop
- [x] support CUDA

### DIFF
--- a/l5kit/l5kit/simulation/dataset.py
+++ b/l5kit/l5kit/simulation/dataset.py
@@ -163,7 +163,7 @@ class SimulationDataset(Dataset):
         self._update_agent_infos(scene_index, frame_agents["track_id"])
         return agents_dict
 
-    def _update_agent_infos(self, scene_index: int, agent_track_ids: np.ndarray):
+    def _update_agent_infos(self, scene_index: int, agent_track_ids: np.ndarray) -> None:
         """Update tracked agents object such that:
         - if agent was not there -> add it
         - if agent is not here anymore -> remove it
--- a/l5kit/l5kit/simulation/dataset.py
+++ b/l5kit/l5kit/simulation/dataset.py
@@ -2,7 +2,6 @@ from copy import deepcopy
 from typing import Dict, List, Set, Tuple
 
 import numpy as np
-from torch.utils.data import Dataset
 
 from l5kit.data import filter_agents_by_frames, PERCEPTION_LABEL_TO_INDEX
 from l5kit.dataset import EgoDataset
@@ -10,7 +9,7 @@ from l5kit.geometry.transform import yaw_as_rotation33
 from l5kit.simulation.utils import disable_agents, insert_agent
 
 
-class SimulationDataset(Dataset):
+class SimulationDataset:
     def __init__(self, dataset: EgoDataset, scene_indices: List[int], start_frame: int = 0,
                  disable_new_agents: bool = False, distance_th_far: float = 15,
                  distance_th_close: float = 10) -> None:
@@ -40,6 +39,8 @@ class SimulationDataset(Dataset):
             scene_dataset = dataset.get_scene_dataset(scene_idx)
             self.scene_dataset_batch[scene_idx] = scene_dataset
 
+        self._len = min([len(scene_dt.dataset.frames) for scene_dt in self.scene_dataset_batch.values()])
+
         # keep track of original dataset
         self.recorded_scene_dataset_batch = deepcopy(self.scene_dataset_batch)
 
@@ -67,19 +68,7 @@ class SimulationDataset(Dataset):
 
         :return: the number of frames
         """
-        num_frame_batch = [len(scene_dt.dataset.frames) for scene_dt in self.scene_dataset_batch.values()]
-        return min(num_frame_batch)
-
-    def __getitem__(self, indices: Tuple[int, int]) -> Dict[str, np.ndarray]:
-        """
-        Get a single frame from a single scene
-
-        :param indices: tuple [scene_idx, frame_idx]
-        :return: a dict from EgoDataset
-        """
-        scene_index, frame_index = indices
-        data_batch = self.scene_dataset_batch[scene_index][frame_index]
-        return data_batch
+        return self._len
 
     def rasterise_frame_batch(self, state_index: int) -> List[Dict[str, np.ndarray]]:
         """
--- a/l5kit/l5kit/simulation/dataset.py
+++ b/l5kit/l5kit/simulation/dataset.py
@@ -90,8 +90,8 @@ class SimulationDataset(Dataset):
         frame_batch = [scene_dt[state_index] for scene_dt in self.scene_dataset_batch.values()]
         return frame_batch
 
-    def set_ego_for_frame(self, state_index: int, output_index: int, ego_translations: np.ndarray,
-                          ego_yaws: np.ndarray) -> None:
+    def set_ego(self, state_index: int, output_index: int, ego_translations: np.ndarray,
+                ego_yaws: np.ndarray) -> None:
         """Mutate future frame position and yaw for ego across scenes. This acts on the underlying dataset
 
         :param state_index: the frame index to mutate
--- a/l5kit/l5kit/simulation/dataset.py
+++ b/l5kit/l5kit/simulation/dataset.py
@@ -1,5 +1,5 @@
 from copy import deepcopy
-from typing import Dict, List, Set, Tuple
+from typing import Dict, List, NamedTuple, Optional, Set, Tuple
 
 import numpy as np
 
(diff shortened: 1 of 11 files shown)


l5kit/l5kit/simulation/dataset.py
```diff
@@ -0,0 +1,216 @@
...
+                 disable_new_agents: bool = False, distance_th_far: float = 15,
+                 distance_th_close: float = 10) -> None:
+        """This class allows to:
+        - rasterise the same frame across multiple scenes for ego;
+        - rasterise the same frame across multiple scenes for multiple agents;
+        - filter agents based on distance to ego;
+        - set ego in future frames;
+        - set agents in future frames;
+
+        NOTE: only vehicles (car label) are picked as agents
```
perone, line 23:
```suggestion
        .. note:: only vehicles (car label) are picked as agents
```

l5kit/l5kit/simulation/utils.py
```diff
@@ -0,0 +1,76 @@
+import numpy as np
+
+from l5kit.data import ChunkedDataset, get_agents_slice_from_frames
+
+
+def insert_agent(agent: np.ndarray, frame_idx: int, dataset: ChunkedDataset):
```
perone, line 6: Given that you're changing the `dataset (ChunkedDataset)`, would it make more sense to move this to be a method on the `ChunkedDataset` or to a wrapper around it ? Just to move it outside from the `utils` module.

l5kit/l5kit/simulation/utils.py
```diff
@@ -0,0 +1,76 @@
...
+        dataset.agents = np.concatenate(
+            [dataset.agents[0: agents_slice.stop], agent, dataset.agents[agents_slice.stop:]], 0
+        )
+
+        # move end of the current frame and all other frames start and end
+        dataset.frames[frame_idx]["agent_index_interval"] += (0, 1)
+        dataset.frames[frame_idx + 1:]["agent_index_interval"] += 1
+
+
+def disable_agents(dataset: ChunkedDataset, allowlist: np.ndarray):
```
perone, line 50: Same here, we're changing the `ChunkedDataset`

perone (dismissed):
Some minor comments. PS: I'm not that familiar w/ the simnet so I will schedule a meeting this week to understand it better.

l5kit/l5kit/simulation/utils.py
lucabergamini: it might be, but I'm worried it will accumulate a lot of methods that way...Also, this only works if the dataset is in memory, so it won't work for a default `ChunkedDataset`. This is probably the wrapper we should have

l5kit/l5kit/simulation/unroll.py
```diff
@@ -0,0 +1,184 @@
...
+    def get_scene_id(self) -> int:
+        """
+        Get the scene index for this SimulationOutputs
+
+        :return: the scene index
+        """
+        return self.scene_id
+
+
+class SimulationLoop:
```
perone, line 63: I would try to keep the same naming we have, like the `ClosedLoopSimulator`.

l5kit/l5kit/simulation/dataset.py
```diff
@@ -0,0 +1,217 @@
...
+        :param distance_th_close: the distance threshold for already tracked agents,
+            An agent will be grabbed if it's closer than this value to ego, defaults to 10
+        """
+        self.scene_indices = scene_indices
+        self.filter_agents_thr = dataset.cfg["raster_params"]["filter_agents_threshold"]
+
+        self.scene_dataset_batch: Dict[int, EgoDataset] = {}  # dicts preserve insertion order
+        for scene_idx in self.scene_indices:
+            scene_dataset = dataset.get_scene_dataset(scene_idx)
+            self.scene_dataset_batch[scene_idx] = scene_dataset
```
perone, line 41: I think it would be better to have the `SimulationDataset` to accept a parameter of the type `scene_dataset_batch: Dict[int, EgoDataset]` and this loop is done by the caller of this class.

l5kit/l5kit/simulation/dataset.py
```diff
@@ -0,0 +1,217 @@
...
+                disable_agents(dataset_zarr, allowlist=frame_agents["track_id"])
+
+    def __len__(self) -> int:
+        """
+        Return the minimum number of frames across scenes
+
+        :return: the number of frames
+        """
+        num_frame_batch = [len(scene_dt.dataset.frames) for scene_dt in self.scene_dataset_batch.values()]
+        return min(num_frame_batch)
```
perone, line 71: Is this going to change or be constant ? If it doesn't change, we can move this to the constructor and just return it here to avoid this counting at every `len()` call.

l5kit/l5kit/simulation/dataset.py
```diff
@@ -0,0 +1,217 @@
...
+
+        This is to avoid acquiring and releasing the same agents if it is on the boundary of the selection
+
+        :param scene_idx: the scene index (used to check for agents_infos)
+        :param frame_agents: the agents in this frame
+        :param ego_pos: the ego position in this frame
+        :return: the filtered agents
+        """
+        # keep only vehicles
+        vehicle_mask = frame_agents["label_probabilities"][:, PERCEPTION_LABEL_TO_INDEX["PERCEPTION_LABEL_CAR"]]
```
perone, line 200:
```suggestion
        car_index = PERCEPTION_LABEL_TO_INDEX["PERCEPTION_LABEL_CAR"]
        vehicle_mask = frame_agents["label_probabilities"][:, car_index]
```

perone (approved):
Just some minor comments.

l5kit/l5kit/simulation/dataset.py
lucabergamini: I'm not sure about this point, as I would need to add a function to perform the same code and call it everywhere...is this buying us anything?

perone: It is mainly to separate concerns. Another alternative would be to have the constructor accepting the `scene_dataset_batch: Dict[int, EgoDataset]` as I mentioned, and a static factory method that would be called something like: `SimulationDataset.from_dataset_indices(dataset, indices, ...)` that will return an instance of the `SimulationDataset` after filtering (that static method will do the filtering and call the constructor). This is mainly to separate how the filtering is done, because otherwise there is no way to change it without overriding the constructor, which violates the open/closed principle that classes should be open for extension but closed for changes. If that requires too many changes, then another alternative also would be to just add the filtering code into a separate method, so that can be changed easily later without having to change the class itself.

lucabergamini: ok, I'll add the static factory method :) thanks for the clarification

l5kit/l5kit/simulation/unroll.py
```diff
@@ -0,0 +1,162 @@
...
+from torch.utils.data.dataloader import default_collate
+from tqdm.auto import tqdm
+
+from l5kit.data import AGENT_DTYPE, PERCEPTION_LABEL_TO_INDEX
+from l5kit.dataset import EgoDataset
+from l5kit.geometry import transform_points
+from l5kit.simulation.dataset import SimulationConfig, SimulationDataset
+
+
+class SimulationOutputs:
```
perone, line 14:
```suggestion
class SimulationOutput:
```

l5kit/l5kit/simulation/unroll.py
```diff
@@ -0,0 +1,162 @@
...
+        if not sim_cfg.use_agents_gt and model_agents is None:
+            raise ValueError("agents model should not be None when simulating agent")
+
+        self.model_ego = torch.nn.Sequential().to(device) if model_ego is None else model_ego.to(device)
+        self.model_agents = torch.nn.Sequential().to(device) if model_agents is None else model_agents.to(device)
+
+        self.device = device
+        self.dataset = dataset
+
+    def unroll(self, scene_indices: List[int]) -> List[SimulationOutputs]:
```
perone, line 68:
```suggestion
    def unroll(self, scene_indices: List[int]) -> List[SimulationOutput]:
```

l5kit/l5kit/simulation/unroll.py
```diff
@@ -0,0 +1,162 @@
...
+            # EGO
+            if not self.sim_cfg.use_ego_gt:
+                ego_input = sim_dataset.rasterise_frame_batch(frame_index)
+                ego_input_dict = default_collate(ego_input)
+                ego_input_dict = {k: v.to(self.device) for k, v in ego_input_dict.items()}
+                ego_output_dict = self.model_ego(ego_input_dict)
+                if should_update:
+                    self.update_ego(sim_dataset, next_frame_index, ego_input_dict, ego_output_dict)
+
+        simulated_outputs: List[SimulationOutputs] = []
```
perone, line 105:
```suggestion
        simulated_outputs: List[SimulationOutput] = []
```

l5kit/l5kit/simulation/unroll.py
```diff
@@ -0,0 +1,162 @@
...
+        for scene_idx in scene_indices:
+            simulated_outputs.append(SimulationOutputs(scene_idx, sim_dataset))
```
perone, line 107:
```suggestion
            simulated_outputs.append(SimulationOutput(scene_idx, sim_dataset))
```

l5kit/l5kit/simulation/unroll.py
```diff
@@ -0,0 +1,162 @@
...
+        self.simulated_dataset = sim_dataset.scene_dataset_batch[scene_id]
+
+        self.recorded_ego_states = self.recorded_dataset.dataset.frames
+        self.recorded_agents_states = self.recorded_dataset.dataset.agents
+        self.simulated_ego_states = self.simulated_dataset.dataset.frames
+        self.simulated_agents_states = self.simulated_dataset.dataset.agents
+
+    def get_scene_id(self) -> int:
+        """
+        Get the scene index for this SimulationOutputs
```
perone, line 36:
```suggestion
        Get the scene index for this SimulationOutput
```

perone (approved):
I just changed all places I found to rename `SimulationOutputs` to `SimulationOutput`, it is something I think we should depart from avjadoo as it is sometimes a little confusing because it implies it has multiple scenes on a simulation output.

perone (approved):

PR #254, opened 2021-04-01
[AVR-972] Add simulation loop

lucabergamini:
Add simulation loop to simulate agents and ego

TODO:

- [ ] define and implement a contract for models output
- [x] return zarrs before and after simulation (for visualisation later)
- [x] return simulation output
- [x] add support for end frame (better handle simulation config in general)
- [x] add e2e tests
- [ ] clean code


l5kit/l5kit/simulation/unroll.py
```diff
@@ -0,0 +1,145 @@
...
+    for frame_index in tqdm(range(start_frame_index, end_frame_index)):
+        next_frame_index = frame_index + 1
+
+        # AGENTS
+        agents_input = dataset.rasterise_agents_frame_batch(frame_index)
+        agents_input_dict = default_collate(list(agents_input.values()))
+        agents_output_dict = model_agents(agents_input_dict)
+
+        # EGO
+        ego_input = dataset.rasterise_frame_batch(frame_index)
```
yawei-ye, line 95: I assume we always get `agent_input` and `ego_input`, we can hide `rasterise_agents_frame_batch` & `rasterise_frame_batch` in the `__get_item__` and put the SimulationDataset in a dataloader.

l5kit/l5kit/simulation/unroll.py
```diff
@@ -0,0 +1,145 @@
...
+        ego_input_dict = default_collate(ego_input)
+        ego_output_dict = model_ego(ego_input_dict)
+
+        if next_frame_index != end_frame_index:
+            update_agents(dataset, next_frame_index, agents_input_dict, agents_output_dict)
+            update_ego(dataset, next_frame_index, ego_input_dict, ego_output_dict)
```
yawei-ye, line 101: shall we create a class `ClosedLoopSimulator` as in jadoo? https://github.com/lyft/avjadoo/blob/00348d969ddfdfc153d5cb0279bf6ff8b3688696/avjadoo/data/ddp_scene/sim/closed_loop_simulation.py#L213

l5kit/l5kit/simulation/unroll.py
lucabergamini: it would be a Dataloader with batch size fixed to 1, and we still need the dataset to set the future info...

l5kit/l5kit/simulation/unroll.py
lucabergamini: I'm not sure I'll buy us anything, but let me try

l5kit/l5kit/simulation/dataset.py
```diff
@@ -39,6 +40,9 @@ def __init__(self, dataset: EgoDataset, scene_indices: List[int], start_frame: i
             scene_dataset = dataset.get_scene_dataset(scene_idx)
             self.scene_dataset_batch[scene_idx] = scene_dataset
 
+        # keep track of original dataset
+        self.recorded_scene_dataset_batch = deepcopy(self.scene_dataset_batch)
```
perone, line 44: This can be very expensive operation, do we need to copy everything here or just some part of the `EgoDataset` objects ?

l5kit/l5kit/simulation/unroll.py
```diff
@@ -0,0 +1,226 @@
...
+
+from l5kit.data import AGENT_DTYPE
+from l5kit.dataset import EgoDataset
+from l5kit.geometry import transform_points
+from l5kit.simulation.dataset import SimulationDataset
+
+
+class SimulationConfig(NamedTuple):
+    """ Defines the parameters used for the simulation of ego and agents around it.
+    :param use_ego_gt: whether to use GT annotations for ego instead of model's outputs
```
perone, line 16:
```suggestion
    """ Defines the parameters used for the simulation of ego and agents around it.
    
    :param use_ego_gt: whether to use GT annotations for ego instead of model's outputs
```

l5kit/l5kit/simulation/unroll.py
```diff
@@ -0,0 +1,226 @@
...
+        :param sim_dataset: the simulation dataset
+        """
+        if scene_id not in sim_dataset.scene_indices:
+            raise ValueError(f"scene: {scene_id} not in {sim_dataset.scene_indices}")
+
+        self.scene_id = scene_id
+        self.recorded_dataset = sim_dataset.recorded_scene_dataset_batch[scene_id]
+        self.simulated_dataset = sim_dataset.scene_dataset_batch[scene_id]
+
+        self.recorded_ego_states = self.recorded_dataset.dataset.frames
```
perone, line 48: Would it make sense to add some methods in the recorded_dataset such as `get_frames()` ?

l5kit/l5kit/simulation/unroll.py
```diff
@@ -0,0 +1,226 @@
...
+                                        self.sim_cfg.disable_new_agents, self.sim_cfg.distance_th_far,
+                                        self.sim_cfg.distance_th_close)
+
+        if self.sim_cfg.num_simulation_steps is None:
+            range_unroll = range(self.sim_cfg.start_frame_index, len(sim_dataset))
+        else:
+            end_frame_index = min(len(sim_dataset), self.sim_cfg.num_simulation_steps + self.sim_cfg.start_frame_index)
+            range_unroll = range(self.sim_cfg.start_frame_index, end_frame_index)
+
+        for frame_index in tqdm(range_unroll):
```
perone, line 100: It would be nice if you can add a flag on the config like `show_progress` to enable/disable `tqdm` here.

l5kit/l5kit/simulation/unroll.py
```diff
@@ -0,0 +1,226 @@
...
+            if not self.sim_cfg.use_ego_gt:
+                ego_input = sim_dataset.rasterise_frame_batch(frame_index)
+                ego_input_dict = default_collate(ego_input)
+                ego_output_dict = self.model_ego(ego_input_dict)
+                if should_update:
+                    self.update_ego(sim_dataset, next_frame_index, ego_input_dict, ego_output_dict)
+
+        simulated_outputs: List[SimulationOutputs] = []
+        for scene_idx in scene_indices:
+            simulated_outputs.append(SimulationOutputs(scene_idx, sim_dataset))
```
perone, line 122: I think it would be better here to pass to `SimulationOutputs` only what is needed from the `SimulationDataset` to build it, instead of the entire object and then getting what is needed from inside of the `SimulationOutputs`, otherwise, it makes it difficult to understand what it uses from the dataset just by looking at the constructor.

l5kit/l5kit/simulation/unroll.py
```diff
@@ -0,0 +1,226 @@
...
+        yaw = input_dict["yaw"]
+        pred_trs = transform_points(output_dict["positions"][:, :1], world_from_agent)[:, 0]
+        pred_yaws = yaw + output_dict["yaws"][:, 0]
+
+        next_agents = np.zeros(len(yaw), dtype=AGENT_DTYPE)
+        next_agents["centroid"] = pred_trs
+        next_agents["yaw"] = pred_yaws
+        next_agents["track_id"] = input_dict["track_id"]
+        next_agents["extent"] = input_dict["extent"]
+        next_agents["label_probabilities"][:, 3] = 1
```
perone, line 151: I think we should create some enums for the fields here, I don't really know what is the `3` for instance.

l5kit/l5kit/simulation/unroll.py
```diff
@@ -0,0 +1,226 @@
...
+
+    @staticmethod
+    def update_ego(dataset: SimulationDataset, frame_idx: int, input_dict: Dict[str, torch.Tensor],
+                   output_dict: Dict[str, torch.Tensor]) -> None:
+        """Update ego across scenes for the given frame index.
+
+        :param dataset: The simulation dataset
+        :param frame_idx: index of the frame to modify
+        :param input_dict: the input to the ego model
+        :param output_dict: the output of the ego model
```
perone, line 165: It would be nice to add a doc for the tensor shape here.

l5kit/l5kit/simulation/unroll.py
```diff
@@ -0,0 +1,226 @@
...
+        pred_trs = transform_points(output_dict["positions"][:, :1], world_from_agent)
+        pred_yaws = yaw + output_dict["yaws"][:, :1]
+
+        dataset.set_ego_for_frame(frame_idx, 0, pred_trs, pred_yaws)
+
+
+from l5kit.configs import load_config_data
+from l5kit.data import ChunkedDataset, LocalDataManager
+from l5kit.dataset import EgoDataset
+from l5kit.rasterization import build_rasterizer
```
perone, line 182: Move to the top

l5kit/l5kit/simulation/unroll.py
```diff
@@ -0,0 +1,226 @@
...
+
+
+class MockModel(torch.nn.Module):
```
perone, line 185: Missing documentation.

l5kit/l5kit/simulation/unroll.py
```diff
@@ -0,0 +1,226 @@
...
+
+    sim_cfg = SimulationConfig(use_ego_gt=False,
+                               use_agents_gt=False,
+                               disable_new_agents=False,
+                               distance_th_far=10,
+                               distance_th_close=30,
+                               start_frame_index=0,
+                               num_simulation_steps=20)
+    sim = SimulationLoop(sim_cfg, MockModel(), MockModel())
+    sim.unroll(EgoDataset(cfg, zarr_dt, rast), [5, 6, 7, 8])
```
perone, line 226: I think it is better to move this to the tests.

l5kit/l5kit/tests/simulation/unroll_test.py
```diff
@@ -0,0 +1,103 @@
...
+    def forward(self, x: Dict[str, torch.Tensor]) -> Dict[str, torch.Tensor]:
+        centroids = x["centroid"]
+        bs = len(centroids)
+
+        positions = torch.zeros(bs, 12, 2, device=centroids.device)
+        positions[..., 0] = self.advance_x
+
+        yaws = torch.zeros(bs, 12, device=centroids.device)
+
+        return {"positions": positions, "yaws": yaws}
```
perone, line 33: Looks like it is repeated here, the `MockModel`.

perone (commented):
Thanks, added some comments !

l5kit/l5kit/simulation/dataset.py
lucabergamini: we may shave off things later, but for now it's safer to have access to the whole dataset I would say

l5kit/l5kit/simulation/unroll.py
lucabergamini: which config? the simulation one?

l5kit/l5kit/simulation/unroll.py
lucabergamini: yes, but right now we still don't know what we need, that's the issue :)

l5kit/l5kit/simulation/unroll.py
lucabergamini: this will be removed. it's here only for rapid testing

l5kit/l5kit/simulation/unroll.py
lucabergamini: same as above

l5kit/l5kit/simulation/unroll.py
lucabergamini: same here

l5kit/l5kit/tests/simulation/unroll_test.py
lucabergamini: this will stau

l5kit/l5kit/simulation/unroll.py
lucabergamini: right now it's not P0, but I'm happy to go over this later

l5kit/l5kit/simulation/unroll.py
lucabergamini: I can actually get the index of the label here. We should definitely add some enum later though

PR #255, opened 2021-04-06
[AVR-973] add option to render ego history frames

yawei-ye:
# Changes
- add option to render ego history frames
- change doc style from Google to reStructured


lucabergamini (approved):
LGTM

PR #259, opened 2021-04-09
Add evaluation plan for the closed-loop evaluation

perone:
This PR will add the evaluation plan (`EvaluationPlan`) and its unit-testing as well as composite metrics and validators protocols.

### DIFF
--- a/l5kit/l5kit/cle/closed_loop_evaluator.py
+++ b/l5kit/l5kit/cle/closed_loop_evaluator.py
@@ -1,0 +1,199 @@
+from typing import Dict, Iterable, List, Optional
+
+import torch
+
+from l5kit.cle.composite_metrics import SupportsCompositeMetricCompute
+from l5kit.cle.metrics import SupportsMetricCompute
+from l5kit.cle.validators import SupportsMetricValidate, ValidatorOutput
+from l5kit.simulation.unroll import SimulationOutput
+
+
+class EvaluationPlan:
+    """Evaluation plan describes a plan to evaluate metrics and run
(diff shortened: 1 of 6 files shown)


perone:
PTAL @lucabergamini, I fixed the Py 3.7 issues.

l5kit/setup.py
```diff
@@ -3,6 +3,7 @@
 
 from l5kit import __version__
 
+
```
lucabergamini, line 6: isort should not run on this file tbh but I'm fine with this change

l5kit/l5kit/cle/closed_loop_evaluator.py
```diff
@@ -0,0 +1,199 @@
...
+
+import torch
+
+from l5kit.cle.composite_metrics import SupportsCompositeMetricCompute
+from l5kit.cle.metrics import SupportsMetricCompute
+from l5kit.cle.validators import SupportsMetricValidate, ValidatorOutput
+from l5kit.simulation.unroll import SimulationOutput
+
+
+class EvaluationPlan:
```
lucabergamini, line 11: just to check: this doesn't have a method to:
- evaluate metrics
- evaluate validators
- evaluate composed metrics
 right?

that will come later somewhere else

lucabergamini (approved):
LGTM

l5kit/l5kit/cle/closed_loop_evaluator.py
perone: Correct, that will come from the `ClosedLoopEvaluator`

PR #260, opened 2021-04-09
[AVR-975] Fix SimulationOutput

lucabergamini:
add torch states to `SimulationOutput` which will be used by the evaluation plan. 
Add test for that

NOTE: there is a conversion from float64 to float32 which reduces precision for the torch states

### DIFF
--- a/l5kit/l5kit/simulation/unroll.py
+++ b/l5kit/l5kit/simulation/unroll.py
@@ -1,3 +1,4 @@
+from enum import IntEnum
 from typing import Dict, List, Optional, Tuple
 
 import numpy as np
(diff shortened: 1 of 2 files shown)


l5kit/l5kit/simulation/unroll.py
```diff
@@ -39,6 +63,26 @@ def get_scene_id(self) -> int:
...
         return self.scene_id
 
+    @staticmethod
+    def build_trajectory_states(frames: np.ndarray) -> torch.Tensor:
+        """
+        Convert frames into a torch trajectory
+        :param frames: the scene frames
+        :return: the trajectory
+        """
+        trajectory_states = torch.zeros(len(frames), len(TrajectoryStateIndices.__members__), dtype=torch.float)
```
perone, line 73:
```suggestion
        trajectory_states = torch.zeros(len(frames), len(TrajectoryStateIndices), dtype=torch.float)
```

perone (approved):
Just one minor suggestion.

l5kit/l5kit/simulation/unroll.py
lucabergamini: my IDE was complaining about this, but it works indeed :)

perone (approved):

PR #263, opened 2021-04-12
[AVR-976] Fix simulation source scenes

lucabergamini:
Right now we're not cutting the datasets to the range which will be simulated. This may introduce issues (e.g. in visualisation and metrics). This PR address this

- [x] add util
- [x] test util
- [x] use util in simulation
- [x] add specific test for unroll loop

### DIFF
--- a/l5kit/l5kit/simulation/dataset.py
+++ b/l5kit/l5kit/simulation/dataset.py
@@ -19,7 +19,7 @@ class SimulationConfig(NamedTuple):
     :param distance_th_close: if a new agent is closer than this value to ego, it will be controlled
     :param start_frame_index: the start index of the simulation
     :param num_simulation_steps: the number of step to simulate
-    :param disable_tqdm: whether to disable tqdm during unroll
+    :param show_info: whether to show info logging during unroll
     """
     use_ego_gt: bool
     use_agents_gt: bool
@@ -28,7 +28,7 @@ class SimulationConfig(NamedTuple):
     distance_th_close: float
     start_frame_index: int = 0
     num_simulation_steps: Optional[int] = None
-    disable_tqdm: bool = True
+    show_info: bool = False
 
 
 class SimulationDataset:
--- a/l5kit/l5kit/simulation/dataset.py
+++ b/l5kit/l5kit/simulation/dataset.py
@@ -19,6 +19,7 @@ class SimulationConfig(NamedTuple):
     :param distance_th_close: if a new agent is closer than this value to ego, it will be controlled
     :param start_frame_index: the start index of the simulation
     :param num_simulation_steps: the number of step to simulate
+    :param disable_tqdm: whether to disable tqdm during unroll
     """
     use_ego_gt: bool
     use_agents_gt: bool
@@ -27,6 +28,7 @@ class SimulationConfig(NamedTuple):
     distance_th_close: float
     start_frame_index: int = 0
     num_simulation_steps: Optional[int] = None
+    disable_tqdm: bool = True
 
 
 class SimulationDataset:
@@ -52,9 +54,11 @@ class SimulationDataset:
         # we cut each scene so that it starts from there and ends after `num_simulation_steps`
         start_frame_idx = self.sim_cfg.start_frame_index
         if self.sim_cfg.num_simulation_steps is None:
-            end_frame_idx = min([len(scene_dt.dataset.frames) for scene_dt in self.scene_dataset_batch.values()])
+            end_frame_idx = self.get_min_len()
         else:
             end_frame_idx = start_frame_idx + self.sim_cfg.num_simulation_steps
+            if end_frame_idx > self.get_min_len():
+                raise ValueError(f"can't unroll until frame {end_frame_idx}, length is {self.get_min_len()}")
 
         for scene_idx in scene_dataset_batch:
             zarr_dt = self.scene_dataset_batch[scene_idx].dataset
@@ -63,8 +67,6 @@ class SimulationDataset:
             # this is the only stateful field we need to change for EgoDataset, it's used in bisect
             frame_index_ends = self.scene_dataset_batch[scene_idx].dataset.scenes["frame_index_interval"][:, 1]
             self.scene_dataset_batch[scene_idx].cumulative_sizes = frame_index_ends
-
-        self._len = min([len(scene_dt.dataset.frames) for scene_dt in self.scene_dataset_batch.values()])
 
         # buffer used to keep track of tracked agents during unroll as tuples of scene_idx, agent_idx
         self._agents_tracked: Set[Tuple[int, int]] = set()
@@ -103,13 +105,20 @@ class SimulationDataset:
             scene_dataset_batch[scene_idx] = scene_dataset
         return SimulationDataset(scene_dataset_batch, sim_cfg)
 
+    def get_min_len(self) -> int:
+        """Return the minimum number of frames between the scenes
+
+        :return: the minimum number of frames
+        """
+        return min([len(scene_dt.dataset.frames) for scene_dt in self.scene_dataset_batch.values()])
+
     def __len__(self) -> int:
         """
         Return the minimum number of frames across scenes
 
         :return: the number of frames
         """
-        return self._len
+        return self.get_min_len()
 
     def rasterise_frame_batch(self, state_index: int) -> List[Dict[str, np.ndarray]]:
         """
--- a/l5kit/l5kit/simulation/dataset.py
+++ b/l5kit/l5kit/simulation/dataset.py
@@ -6,7 +6,7 @@ import numpy as np
 from l5kit.data import filter_agents_by_frames, PERCEPTION_LABEL_TO_INDEX
 from l5kit.dataset import EgoDataset
 from l5kit.geometry.transform import yaw_as_rotation33
-from l5kit.simulation.utils import disable_agents, insert_agent
+from l5kit.simulation.utils import disable_agents, get_frames_subset, insert_agent
 
 
 class SimulationConfig(NamedTuple):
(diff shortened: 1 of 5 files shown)


l5kit/l5kit/simulation/dataset.py
```diff
@@ -46,19 +46,34 @@ def __init__(self, scene_dataset_batch: Dict[int, EgoDataset], sim_cfg: Simulati
...
         self.scene_dataset_batch: Dict[int, EgoDataset] = scene_dataset_batch
+        self.sim_cfg = sim_cfg
 
-        self._len = min([len(scene_dt.dataset.frames) for scene_dt in self.scene_dataset_batch.values()])
+        # we must limit the scenes to the part which will be simulated
+        # we cut each scene so that it starts from there and ends after `num_simulation_steps`
+        start_frame_idx = self.sim_cfg.start_frame_index
+        if self.sim_cfg.num_simulation_steps is None:
+            end_frame_idx = min([len(scene_dt.dataset.frames) for scene_dt in self.scene_dataset_batch.values()])
+        else:
```
perone, line 56: Do we want to support when the number of simulation steps is greater than the minimum length ? I think it would be better to add a condition and raise an exception for now.

l5kit/l5kit/simulation/dataset.py
```diff
@@ -46,19 +46,34 @@ def __init__(self, scene_dataset_batch: Dict[int, EgoDataset], sim_cfg: Simulati
...
-        self.sim_cfg = sim_cfg
+        for scene_idx in scene_dataset_batch:
+            zarr_dt = self.scene_dataset_batch[scene_idx].dataset
+            self.scene_dataset_batch[scene_idx].dataset = get_frames_subset(zarr_dt, start_frame_idx, end_frame_idx)
+
+            # this is the only stateful field we need to change for EgoDataset, it's used in bisect
+            frame_index_ends = self.scene_dataset_batch[scene_idx].dataset.scenes["frame_index_interval"][:, 1]
+            self.scene_dataset_batch[scene_idx].cumulative_sizes = frame_index_ends
+
+        self._len = min([len(scene_dt.dataset.frames) for scene_dt in self.scene_dataset_batch.values()])
```
perone, line 67: This code is repeating the same as line 55 above. You can create a method for it.

l5kit/l5kit/simulation/unroll.py
```diff
@@ -117,15 +117,9 @@ def unroll(self, scene_indices: List[int]) -> List[SimulationOutput]:
...
         sim_dataset = SimulationDataset.from_dataset_indices(self.dataset, scene_indices, self.sim_cfg)
 
-        if self.sim_cfg.num_simulation_steps is None:
-            range_unroll = range(self.sim_cfg.start_frame_index, len(sim_dataset))
-        else:
-            end_frame_index = min(len(sim_dataset), self.sim_cfg.num_simulation_steps + self.sim_cfg.start_frame_index)
-            range_unroll = range(self.sim_cfg.start_frame_index, end_frame_index)
-
-        for frame_index in tqdm(range_unroll):
+        for frame_index in tqdm(range(len(sim_dataset))):
```
perone, line 120: I would add a config like `show_progress` in case the user wants to hide the output, `tqdm` has an option called `disable` that is often used for that.

l5kit/l5kit/simulation/utils.py
```diff
@@ -74,3 +74,50 @@ def disable_agents(dataset: ChunkedDataset, allowlist: np.ndarray) -> None:
     dataset.agents["yaw"][mask_disable] *= 0
     dataset.agents["extent"][mask_disable] *= 0
     dataset.agents["label_probabilities"][mask_disable] = -1
+
+
+def get_frames_subset(dataset: ChunkedDataset, frame_start_idx: int, frame_end_idx: int) -> ChunkedDataset:
+    """Get a new dataset with frames between start (included) and end (excluded).
+    Assumptions:
+        the dataset has only 1 scene
+        the dataset is in numpy format and not zarr anymore
```
perone, line 83: I think this will be rendered all together in Sphinx (not sure though)

l5kit/l5kit/simulation/utils.py
```diff
@@ -74,3 +74,50 @@ def disable_agents(dataset: ChunkedDataset, allowlist: np.ndarray) -> None:
...
+
+    """
+    if not len(dataset.scenes) == 1:
+        raise ValueError(f"dataset should have a single scene, got {len(dataset.scenes)}")
+    if not isinstance(dataset.agents, np.ndarray):
+        raise ValueError("dataset agents should be an editable np array")
+    if not isinstance(dataset.tl_faces, np.ndarray):
+        raise ValueError("dataset tls should be an editable np array")
+    if not isinstance(dataset.frames, np.ndarray):
+        raise ValueError("dataset frames should be an editable np array")
```
perone, line 97: I would remove that for two reasons: the NumPy array can be non-writeable and we are not checking this here, and if a different type is provided we will see clear errors. Also, if we want to do runtime type-checking, I would go with another solution that uses the annotations for that automatically. So I would just remove the check for the type.

l5kit/l5kit/simulation/utils.py
```diff
@@ -74,3 +74,50 @@ def disable_agents(dataset: ChunkedDataset, allowlist: np.ndarray) -> None:
...
+    if frame_start_idx >= len(dataset.frames):
+        raise ValueError(f"frame start {frame_start_idx} is over the length of the dataset")
+    if frame_end_idx > len(dataset.frames):
+        raise ValueError(f"frame end {frame_end_idx} is over the length of the dataset")
+    if frame_start_idx >= frame_end_idx:
+        raise ValueError(f"end frame {frame_end_idx} should be higher than start {frame_start_idx}")
+    if frame_start_idx < 0:
+        raise ValueError(f"start frame {frame_start_idx} should be positive")
+
+    new_dataset = ChunkedDataset("")
```
perone, line 107: Just to understand here, what the `""` mean for the `ChunkedDataset` ?

perone (approved):
Thanks, just some minor comments.

l5kit/l5kit/simulation/dataset.py
lucabergamini: it will raise a `ValueError` later when trying to cut the zarr, but I can add a check here sure

l5kit/l5kit/simulation/dataset.py
lucabergamini: sure

l5kit/l5kit/simulation/unroll.py
lucabergamini: the question is where should we add it so that it's visible here. Is the SImulationConfig the right place?

l5kit/l5kit/simulation/utils.py
lucabergamini: because of the missing return?

l5kit/l5kit/simulation/utils.py
lucabergamini: > I would remove that for two reasons: the NumPy array can be non-writeable and we are not checking this here

but this doesn't happen in this codebase. You would really need to go behind our API to get a non writable numpy array

> Also, if we want to do runtime type-checking, I would go with another solution that uses the annotations for that automatically. So I would just remove the check for the type.

The only way I see is another class similar to `ChunkedDataset` with numpy arrays. But that would affect a lot of code and it's not clear to me how it would interact with the `EgoDataset`..

l5kit/l5kit/simulation/utils.py
lucabergamini: the `ChunkedDataset` has a `path` members which will be used when `open` or `initalise` are called. This is an old logic which we should probably replace.... the argument is required so that's why I'm passing and empty string here

l5kit/l5kit/simulation/unroll.py
perone: Yes, I would add in the `SimulationConfig`.

l5kit/l5kit/simulation/utils.py
perone: Because of the line break, I think you need a `-` in front of it or another line-break.

l5kit/l5kit/simulation/utils.py
perone: I see, let's leave the checks then, but when that would be the case that we don't have for instance a numpy array for the `dataset.tl_faces` ?

l5kit/l5kit/simulation/utils.py
perone: Thanks

l5kit/l5kit/simulation/utils.py
lucabergamini: the default `ChunkedDataset` uses `zarr.arrays` when there is a map with the disk. We have `numpy.array` only after calling functions such as `get_scene_dataset` which get a memory only version of a subset of the original zarr

perone: Ok now I see the problem, it really makes sense to have another dataset or a flag in the dataset saying when it is in-memory or persistent data. But that doesn't make sense to do in this PR. Thanks !

l5kit/l5kit/simulation/dataset.py
```diff
@@ -27,6 +28,7 @@ class SimulationConfig(NamedTuple):
     distance_th_close: float
     start_frame_index: int = 0
     num_simulation_steps: Optional[int] = None
+    disable_tqdm: bool = True
```
perone, line 31: nit: I would change that to `show_progress` or `show_info` just to make it general for print/logging later, otherwise if we use that flag later for logging/other prints we would have to change the naming here.

perone (approved):
Just one minor comment on the new config name.

perone (approved):
💯 🥇 👍🏻

PR #267, opened 2021-04-13
Adding collision metrics for the closed-loop evaluator

perone:
This PR will add the collision metrics, the tests also depend on the custom mocked data.
I'm using here the `simulation_output.recorded_agents` but I'm not sure about what should come in this array, so I'll submit this PR and then when we have data coming in I can change and fix it.

### DIFF
--- a/l5kit/l5kit/cle/metrics.py
+++ b/l5kit/l5kit/cle/metrics.py
@@ -1,11 +1,13 @@
-from abc import abstractmethod
+from abc import ABC, abstractmethod
 
+import numpy as np
 import torch
 from typing_extensions import Protocol
 
 from l5kit.evaluation import error_functions
 from l5kit.evaluation import metrics as l5metrics
-from l5kit.simulation.unroll import SimulationOutput
+from l5kit.rasterization import EGO_EXTENT_LENGTH, EGO_EXTENT_WIDTH
+from l5kit.simulation.unroll import SimulationOutput, TrajectoryStateIndices
 
 
 class SupportsMetricCompute(Protocol):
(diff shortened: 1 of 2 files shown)


l5kit/l5kit/cle/metrics.py
```diff
@@ -23,6 +25,84 @@ def compute(self, simulation_output: SimulationOutput) -> torch.Tensor:
...
+        :param observed_frame: the ground-truth frame
+        :param simulated_frame_ego_state: ego state from the simulated frame,
+                                          this is a 1D array with the frame
+                                          ego state
+        :returns: metric result for the frame, where 1 means a collision,
+                  and 0 otherwise.
+        """
+        simulated_centroid = simulated_frame_ego_state[:TrajectoryStateIndices.Y + 1].cpu().numpy()
+        simulated_angle = simulated_frame_ego_state[TrajectoryStateIndices.THETA].cpu().numpy()
+        simulated_extent = np.r_[EGO_EXTENT_LENGTH, EGO_EXTENT_WIDTH]
```
lucabergamini, line 50: this is the same as `np.asarray([EGO_EXTENT_LENGTH, EGO_EXTENT_WIDTH])` right?

l5kit/l5kit/cle/metrics.py
```diff
@@ -23,6 +25,84 @@ def compute(self, simulation_output: SimulationOutput) -> torch.Tensor:
...
+        simulated_scene_ego_state = simulation_output.simulated_ego_states
+        simulated_agents = simulation_output.simulated_agents
+
+        if len(simulated_agents) < len(simulated_scene_ego_state):
+            raise ValueError("More simulated timesteps than observed.")
+
+        num_frames = simulated_scene_ego_state.size(0)
+        metric_results = torch.zeros(num_frames, device=simulated_scene_ego_state.device)
+        for frame_idx in range(num_frames):
+            recorded_agent_frame = simulated_agents[frame_idx]
```
lucabergamini, line 75: lol naming here could be improved; unless this is here to test if I'm paying attention while reviewing xD

lucabergamini (commented):
1 minor

l5kit/l5kit/cle/metrics.py
perone: Yes, it is just a short hand to look closer to how lists are built in Python.

l5kit/l5kit/cle/metrics.py
perone: haha will fix it

lucabergamini (approved):

PR #274, opened 2021-04-19
Add new function for bbox world coords

lucabergamini:
Add utility function to get world coordinates of bboxes starting from `agents` array. 
Add tests for the function and use it in visualizer

### DIFF
--- a/l5kit/l5kit/rasterization/box_rasterizer.py
+++ b/l5kit/l5kit/rasterization/box_rasterizer.py
@@ -25,6 +25,25 @@ def get_ego_as_agent(frame: np.ndarray) -> np.ndarray:  # TODO this can be usefu
     return ego_agent
 
 
+def get_box_world_coords(agents: np.ndarray) -> np.ndarray:
+    """Get world coordinates of the 4 corners of the bounding boxes
+
+    :param agents: agents array of size N with centroid (world coord), yaw and extent
+    :return: array of shape (N, 4, 2) with the four corners of each agent
+    """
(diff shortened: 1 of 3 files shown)


l5kit/l5kit/rasterization/box_rasterizer.py
```diff
@@ -25,6 +25,25 @@ def get_ego_as_agent(frame: np.ndarray) -> np.ndarray:  # TODO this can be usefu
...
+    corners_base_coords = (np.asarray([[-1, -1], [-1, 1], [1, 1], [1, -1]]) * 0.5)[None, :, :]
+
+    # compute the corner in world-space (start in origin, rotate and then translate)
+    corners_m = corners_base_coords * agents["extent"][:, None, :2]  # corners in zero
+    s = np.sin(agents["yaw"])
+    c = np.cos(agents["yaw"])
+    # note this is clockwise because it's right-multiplied and not left-multiplied later,
+    # and therefore we're still rotating counterclockwise.
+    rotation_m = np.moveaxis(np.array(((c, s), (-s, c))), 2, 0)
+    box_world_coords = corners_m @ rotation_m + agents["centroid"][:, None, :2]
```
perone, line 43: I would add a comment here about what is the `[:, None, :2]` is doing or you can (even better) just add another variable with it, so you can name what is this term here.

perone (dismissed):
Just a minor comment about adding a comment (no pun intended).

l5kit/l5kit/rasterization/box_rasterizer.py
lucabergamini: I've added comments for all the shape where `None` was involved :)

perone (approved):
Thanks !!

PR #277, opened 2021-04-20
Update visualisation notebook

lucabergamini:
Update the visualisation notebook with an example from the new visualizer.

Refactor a function to allow to load `MapAPI` easily


perone (approved):
👍🏻

PR #298, opened 2021-06-10
bugfix in yaw fitting

yawei-ye:
# Changes
* Bugfix in yaw residual compute, it should be wrapped

### DIFF
--- a/l5kit/l5kit/kinematic/ackerman_steering_model.py
+++ b/l5kit/l5kit/kinematic/ackerman_steering_model.py
@@ -70,7 +70,7 @@ Returns:
             [
                 x - gx,
                 y - gy,
-                r - gr,
+                angular_distance(r, gr),
                 v - gv,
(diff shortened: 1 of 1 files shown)


yawei-ye:
PTAL @lucabergamini @perone

perone (dismissed):

lucabergamini (approved):

perone (approved):
LGTM. Do you think it would be easy to add a test case to cover that issue ? If yes then I would say it would be nice to add, but if it would take too much data mocking to cover it then its ok.

yawei-ye:
> LGTM. Do you think it would be easy to add a test case to cover that issue ? If yes then I would say it would be nice to add, but if it would take too much data mocking to cover it then its ok.

Good point, I created a JIRA ticket: https://jira.lyft.net/browse/AVSW-76244

yawei-ye:
💨

PR #306, opened 2021-07-16
L5 CLE gym envt added

theDebugger811:
L5Kit Gym Environment created for closed loop training


l5kit/l5kit/evaluation/error_functions.py
```diff
@@ -15,3 +16,16 @@ def l2_error(pred: torch.Tensor, gt: torch.Tensor) -> torch.Tensor:
     :return: l2 distance between the predicted and gt tensor, size: [batch_size,]
     """
     return torch.norm(pred - gt, p=2, dim=-1)
+
+
+def closest_angle(angle_a: torch.Tensor, angle_b: torch.Tensor) -> torch.Tensor:
+    """ Finds the closest angle between angle_b - angle_a in radians.
+        Returns an angle between [-pi, pi]
```
perone, line 23: Move this phrase to the `:return:` below.

l5kit/l5kit/simulation/unroll.py
```diff
@@ -76,6 +76,15 @@ def __init__(self, scene_id: int, sim_dataset: SimulationDataset,
...
 
+        # Edit
+        # # Required for Bokeh Visualizer
+        # self.tls_frames = self.simulated_dataset.dataset.tl_faces
+        # self.agents_th = self.simulated_dataset.cfg["raster_params"]["filter_agents_threshold"]
+
+        # # Remove Dataset attributes
+        # self.recorded_dataset = None
+        # self.simulated_dataset = None
+
```
perone, line 87: remove commented code

l5kit/l5kit/simulation/unroll.py
```diff
@@ -122,7 +131,8 @@ def __init__(self, sim_cfg: SimulationConfig, dataset: EgoDataset,
         """
         self.sim_cfg = sim_cfg
         if not sim_cfg.use_ego_gt and model_ego is None:
-            raise ValueError("ego model should not be None when simulating ego")
+            print("ego model should not be None when simulating ego")
```
perone, line 134: Is this not an error anymore ?

perone (commented):
Thanks @theDebugger811, I added a few comments, let's sync on the meeting today.

l5kit/Pipfile
```diff
@@ -13,8 +13,9 @@ sphinx = "*"
 sphinx-rtd-theme = "*"
 recommonmark = "*"
 pre-commit = "*"
-importlib-metadata = "*"  # required for python 3.8
 typing-extensions = "*"
+line-profiler = "*"
+importlib-metadata = "*"
```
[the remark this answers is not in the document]
theDebugger811: I use line-profiler to time the code. Is it okay to remove in the final merge? I will keep a note.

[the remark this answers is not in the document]
perone: Leave it only for your development env then, you can remove from here.

l5kit/l5kit/environment/cle_metricset.py
```diff
@@ -0,0 +1,120 @@
...
+        self.evaluator.evaluate(sim_outputs)
+
+    def reset(self) -> None:
+        """Reset the current state of the CLE (Closed-loop Evaluator)."""
+        self.evaluator.reset()
+
+    def get_results(self) -> Dict[str, Any]:
+        """Perform all required aggregations and returns a dictionary
+        composed by all results."""
+        pass
```
perone, line 102: raise a `NotImplementedError` here.

l5kit/l5kit/environment/envs/l5_env.py
```diff
@@ -0,0 +1,200 @@
...
+        Analogous to sim_loop.unroll(scenes_to_unroll) in L5Kit
+
+        :param action: the action to perform on current state/frame
+        :return: the namedTuple comprising the (next observation, reward, done, info)
+                 based on the current action
+        """
+
+        frame_index = self.frame_index
+        next_frame_index = frame_index + 1
+        episode_over = next_frame_index == (len(self.sim_dataset) - self.future_num_frames)  # Note !!
```
perone, line 129: What is note here ?

l5kit/l5kit/environment/envs/l5_env.py
```diff
@@ -0,0 +1,200 @@
...
+                                                       self.ego_ins_outs, self.agents_ins_outs)
+
+        # done is True when episode ends or error gets too high (optional)
+        done = episode_over or self.check_done_status()
+
+        # Optionally we can pass additional info
+        # We are using "info" to output simulated outputs
+        info = {}
+        if done:
+            info = {"info": self.get_simulated_outputs()}
```
perone, line 159: Do we need this ? This will be copied through IPC so it might be very expensive.

l5kit/l5kit/environment/envs/l5_env.py
```diff
@@ -0,0 +1,200 @@
...
+        assert len(self.scene_indices) == 1
+        # generate simulated_outputs
+        simulated_outputs: List[SimulationOutputGym] = []
+        for scene_idx in self.scene_indices:
+            simulated_outputs.append(SimulationOutputGym(scene_idx, self.sim_dataset,
+                                                         self.ego_ins_outs, self.agents_ins_outs))
+        return simulated_outputs
+
+    def render(self) -> None:
+        pass
```
perone, line 197: I think it would be better to raise a `NotImplementError` error here, to be explicit about the support of this.

l5kit/l5kit/environment/envs/l5_env.py
```diff
@@ -0,0 +1,200 @@
...
+
+    def close(self) -> None:
```
perone, line 199: When is this called ?

l5kit/l5kit/environment/reward.py
```diff
@@ -0,0 +1,91 @@
...
+
+import numpy as np
+import torch
+
+from l5kit.simulation.dataset import SimulationDataset
+from l5kit.simulation.unroll import UnrollInputOutput
+from l5kit.environment.cle_metricset import L5BaseMetricSet, SimulationOutputGym
+
+
+class Reward(object):
```
perone, line 11: Please create an abstraction here: you will have `Reward` (that can be an ABC) and then you will have something like a `L2ImitationReward` (this concrete implementation of the `Reward`). Imagine that can exist multiple rewards later.

l5kit/l5kit/environment/reward.py
```diff
@@ -0,0 +1,91 @@
...
+            simulated_outputs.append(SimulationOutputGym(scene_idx, sim_dataset, ego_ins_outs, agents_ins_outs))
+        self.metric_set.evaluate(simulated_outputs)
+
+        # get CLE metrics
+        scene_metrics = self.metric_set.evaluator.scene_metric_results[scene_idx]
+        dist_error = scene_metrics['displacement_error_l2'][frame_index + 1]
+        yaw_error = self.yaw_weight * torch.abs(scene_metrics['yaw_error_ca'][frame_index + 1])
+
+        # clip reward
+        reward = max(-self.rew_clip_thresh, -dist_error.item())
```
perone, line 66: Add an option to enable/disable it.

l5kit/l5kit/environment/test_gym.py
```diff
@@ -0,0 +1,27 @@
+import os
```
perone, line 1: This file should go to the examples folder if it is an example, otherwise should be an integration test.

perone (commented):
Added some more comments, things we already aligned that are missing is the unit-testing and check the linting that is failing due to a line size I think. Thanks a lot @theDebugger811 !

theDebugger811:
Hi @perone, Thanks for the detailed review. I have incorporated the changes. Also, I added SimulationOutputCLE base class (that gets passed to Close-loop evaluator). All the tests are passing locally but not here. Any idea?

examples/RL/test_gym.py
```diff
@@ -0,0 +1,29 @@
+import os
```
perone, line 1: Rename this file from `test_gym.py` to `create_gym_example.py`, files prefixed with `test_` are often captured by testing systems like `pytest`, etc, as being unit-tests that should be executed during testing.

examples/RL/test_gym.py
```diff
@@ -0,0 +1,29 @@
...
+import argparse
+
+import gym
+
+from l5kit.environment.register_l5_env import create_l5_env
+
+parser = argparse.ArgumentParser()
+parser.add_argument('--data_path', type=str, default='/home/ubuntu/level5_data/',
```
perone, line 9: These paths do not exist in users computers.

examples/RL/test_gym.py
```diff
@@ -0,0 +1,29 @@
...
+                    help='Environment configuration file')
+parser.add_argument('--env_config_path', type=str, default='/home/ubuntu/src/l5kit/examples/RL/config.yaml',
```
perone, line 11: These paths do not exist in users computers.

examples/RL/test_gym.py
```diff
@@ -0,0 +1,29 @@
...
+parser.add_argument('--rew_clip', default=15, type=float,
+                    help='Reward Clipping Threshold')
+args = parser.parse_args()
+
+# By setting the L5KIT_DATA_FOLDER variable, we can point the script to the folder where the data lies.
+# The path to 'l5_data_path' needs to be provided
+os.environ["L5KIT_DATA_FOLDER"] = args.data_path
+
+# create and register L5Kit gym env
+create_l5_env(args.env_config_path, args.eps_length, args.disable_cle, args.rew_clip)
```
perone, line 26: Isn't this supposed to be called on a `__init__.py` of l5kit ? Please see [this reference](https://github.com/openai/gym-soccer/blob/master/gym_soccer/__init__.py).

examples/RL/test_gym.py
```diff
@@ -0,0 +1,29 @@
...
+
+# call
+env = gym.make('L5-CLE-v0')
```
perone, line 29: Can you add some calls here to show the environment ? Example: get one initial observation and print the shapes.

l5kit/l5kit/environment/cle_metricset.py
```diff
@@ -0,0 +1,117 @@
...
+        """
+        super(SimulationOutputGym, self).__init__(scene_id, sim_dataset, ego_ins_outs, agents_ins_outs)
+
+        # Required for Bokeh Visualizer
+        simulated_dataset = sim_dataset.scene_dataset_batch[scene_id]
+        self.tls_frames = simulated_dataset.dataset.tl_faces
+        self.agents_th = simulated_dataset.cfg["raster_params"]["filter_agents_threshold"]
+
+
+class CLEMetricSet(ABC):
```
perone, line 30: Rename this class to `CLEBaseMetricSet` as this is the base (abstract base class) class.

l5kit/l5kit/environment/cle_metricset.py
```diff
@@ -0,0 +1,117 @@
...
+        raise NotImplementedError
+
+    @abstractmethod
+    def get_results(self) -> Dict[str, Any]:
+        """Perform all required aggregations and returns a dictionary
+        composed by all results."""
+        raise NotImplementedError
+
+
+class L5BaseMetricSet(CLEMetricSet):
```
perone, line 50: Can remove the `Base` from the name here.

l5kit/l5kit/environment/envs/l5_env.py
```diff
@@ -0,0 +1,185 @@
...
+from l5kit.environment.cle_metricset import SimulationOutputGym
+from l5kit.environment.reward import Reward, RewardInput
+from l5kit.environment.utils import convert_to_dict, default_collate_numpy, rescale_action
+from l5kit.rasterization import build_rasterizer
+from l5kit.simulation.dataset import SimulationConfig, SimulationDataset
+from l5kit.simulation.unroll import ClosedLoopSimulator, UnrollInputOutput
+
+
+class GymStepOutput(NamedTuple):
+    """ The output dict of gym env.step
```
perone, line 22:
```suggestion
    """The output dict of gym env.step
```

l5kit/l5kit/environment/envs/l5_env.py
```diff
@@ -0,0 +1,185 @@
...
+class L5Env(gym.Env):
+    """ Custom Environment of L5 Kit that can be registered in OpenAI Gym. """
+
+    def __init__(self, env_config_path: str, sim_cfg: SimulationConfig, reward: Reward, cle: bool) -> None:
+        """
+        :param env_config_path: path to the L5Kit environment configuration file
+        :param simulation_cfg: configuration of the L5Kit closed loop simulator
+        :param reward: calculates the reward for the gym environment
+        :param cle: flag to enable close loop environment updates
+        """
```
perone, line 45: This docstring should to together with the one above (Custom Environment of ...)

l5kit/l5kit/environment/envs/l5_env.py
```diff
@@ -0,0 +1,185 @@
...
+    """
+
+    obs: Dict[str, np.ndarray]
+    reward: float
+    done: bool
+    info: Dict[str, Any]
+
+
+class L5Env(gym.Env):
+    """ Custom Environment of L5 Kit that can be registered in OpenAI Gym. """
```
perone, line 37:
```suggestion
    """Custom Environment of L5 Kit that can be registered in OpenAI Gym. """
```

l5kit/l5kit/environment/envs/l5_env.py
```diff
@@ -0,0 +1,185 @@
...
+
+        # init dataset
+        train_zarr = ChunkedDataset(dm.require(cfg["train_data_loader"]["key"])).open()
+        self.dataset = EgoDataset(cfg, train_zarr, rasterizer)
+
+        self.future_num_frames = cfg["model_params"]["future_num_frames"]
+
+        # Define action and observation space
+        # Continuous Action Space: gym.spaces.Box (X, Y, Yaw * number of future states)
+        # self.action_space = spaces.Box(low=-1000, high=1000, shape=(self.future_num_frames * 3, ))
```
perone, line 65: Old commented code.

l5kit/l5kit/environment/envs/l5_env.py
```diff
@@ -0,0 +1,185 @@
...
+        self.sim_cfg = sim_cfg
+        self.simulator = ClosedLoopSimulator(self.sim_cfg, self.dataset, device=torch.device("cpu"),
+                                             verify_model=False)
+
+        # Reward
+        self.reward = reward
+
+        # Scene Sampling config
+        self.max_scene_id = cfg["gym_params"]["max_scene_id"]
+        if cfg["gym_params"]["overfit"]:  # Overfit
```
perone, line 82: This comment isn't adding much here, can remove.

l5kit/l5kit/environment/envs/l5_env.py
```diff
@@ -0,0 +1,185 @@
...
+        """
+
+        frame_index = self.frame_index
+        next_frame_index = frame_index + 1
+        episode_over = next_frame_index == (len(self.sim_dataset) - self.future_num_frames)
+
+        # EGO
+        if not self.sim_cfg.use_ego_gt:
+            action = rescale_action(action)
+            action = convert_to_dict(action, self.future_num_frames)
```
perone, line 130: These two functions `rescale_action` and `convert_to_dict` are very particular of the environment, can you move these functions to be methods of the environment ? They can be private prefixed w/ `_`, as in `def _rescale_action(...)`.

examples/RL/test_gym.py
perone: Note that there are some parameters here that should be supported in the `gym.make("L5-CLE-v0", [args])` args, for instance, if the user wants to change the episode length after registration. Reference for the `make()` -> https://github.com/openai/gym/blob/0.10.10/gym/envs/registration.py#L77.

l5kit/l5kit/environment/reward.py
```diff
@@ -0,0 +1,141 @@
...
+import numpy as np
+import torch
+
+from l5kit.environment.cle_metricset import L5BaseMetricSet, L5GymCLEMetricSet, SimulationOutputCLE
+from l5kit.simulation.dataset import SimulationDataset
+from l5kit.simulation.unroll import UnrollInputOutput
+
+
+class RewardInput(NamedTuple):
+    """ The input tuple to calculate reward
```
perone, line 13:
```suggestion
    """The input tuple to calculate reward
```

Remove the space in the beginning of the docstrings. [Reference for example](https://sphinx-rtd-tutorial.readthedocs.io/en/latest/docstrings.html#an-example-class-with-docstrings).

l5kit/l5kit/environment/utils.py
```diff
@@ -0,0 +1,87 @@
...
+
+import numpy as np
+import torch
+from PIL import Image
+
+from l5kit.rasterization import Rasterizer
+
+
+def rescale_action(action: np.ndarray, x_mu: float = 1.20, x_scale: float = 0.2,
+                   y_mu: float = 0.0, y_scale: float = 0.03, yaw_scale: float = 3.14) -> np.ndarray:
```
perone, line 12: Missing documentation.

l5kit/l5kit/evaluation/error_functions.py
```diff
@@ -15,3 +16,16 @@ def l2_error(pred: torch.Tensor, gt: torch.Tensor) -> torch.Tensor:
     :return: l2 distance between the predicted and gt tensor, size: [batch_size,]
     """
     return torch.norm(pred - gt, p=2, dim=-1)
+
+
+def closest_angle(angle_a: torch.Tensor, angle_b: torch.Tensor) -> torch.Tensor:
+    """ Finds the closest angle between angle_b - angle_a in radians.
```
perone, line 22:
```suggestion
    """Finds the closest angle between angle_b - angle_a in radians.
```

perone (commented):
Thanks @theDebugger811, added some more comments. After that we need to add the unit-testing.

theDebugger811:
@szagoruyko  @lucabergamini , could you please review this L5Kit gym-compatible environment PR?
Thanks!

l5kit/l5kit/cle/metric_set.py
```diff
@@ -0,0 +1,86 @@
...
+        raise NotImplementedError
+
+    @abstractmethod
+    def get_results(self) -> Dict[str, Any]:
+        """Perform all required aggregations and returns a dictionary
+        composed by all results."""
+        raise NotImplementedError
+
+
+class L5MetricSet(BaseMetricSet):
```
lucabergamini, line 30: shouldn't this be doing something in the methods?

l5kit/l5kit/simulation/unroll.py
```diff
@@ -105,12 +105,33 @@ def build_trajectory_states(frames: np.ndarray) -> torch.Tensor:
         return trajectory_states
 
 
+class SimulationOutput(SimulationOutputCLE):
```
lucabergamini, line 108: I think this will break some visualisation utils (e.g. `simulation_out_to_visualizer_scene`)

l5kit/l5kit/simulation/unroll.py
```diff
@@ -119,11 +140,12 @@ def __init__(self, sim_cfg: SimulationConfig, dataset: EgoDataset,
         :param model_ego: the model to be used for ego
         :param model_agents: the model to be used for agents
         :param keys_to_exclude: keys to exclude from input/output (e.g. huge blobs)
+        :param verify_model: flag to verify the ego/agent model when use_ego_gt/use_agent_gt is False
```
lucabergamini, line 143: Why is this required?

l5kit/l5kit/tests/conftest.py
```diff
@@ -36,6 +36,19 @@ def cfg() -> dict:
     return load_config_data("./l5kit/tests/artefacts/config.yaml")
 
 
+@pytest.fixture(scope="function")
```
lucabergamini, line 39: does this need to be function? Isn't enough once per session?

lucabergamini (commented):
left some comments

l5kit/l5kit/environment/envs/l5_env.py
```diff
@@ -0,0 +1,260 @@
...
+    def __new__(cls, eps_length: int = 32) -> 'SimulationConfigGym':
+        """Constructor method
+        """
+        self = super(SimulationConfigGym, cls).__new__(cls, use_ego_gt=False, use_agents_gt=True,
+                                                       disable_new_agents=False, distance_th_far=30,
+                                                       distance_th_close=15, num_simulation_steps=eps_length)
+        return self
+
+
+class SimulationOutputGym(SimulationOutputCLE):
```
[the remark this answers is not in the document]
theDebugger811: SimulationOutputCLE: Is used for close loop evaluation.
SimulationOutputGym: Is used for visualization.

SimulationOutputCLE contains parameters that suffice for Close loop evaluation.  SimulationOutputGym contains additional parameters that are compatible for Bokeh Visualizer as well. Does this help explain?

l5kit/l5kit/cle/metric_set.py
theDebugger811: I did not understand. Should I shift L5MetricSet class to the 'environment' folder?

l5kit/l5kit/simulation/unroll.py
theDebugger811: The refactored SimulationOutput class has the same attributes as the original SimulationOutput class.  The tests are running fine. Would you suggest to check something regarding breaking visualization utils? 

Christian suggested to make SimulationOutput inherent from SimulationOutputCLE so that close-loop evaluator-specific components are kept separated from visualization components. Let me know if there is a better way to do this.

l5kit/l5kit/simulation/unroll.py
theDebugger811: The ClosedLoopSimulator requires both the SimulationDataset and an ego_model when use_ego_gt = False. However, in gym environment, the SimulationDataset (part of environment) is independent of the ego_model (the policy). So, the ClosedLoopSimulator should not have an ego_model (it is not part of environment), and this will raise an error. 

The verify_model flag helps to check whether we want to verify the presence of ego_model (as in L5Kit) or not (for gym). Does this explanation help?

l5kit/l5kit/environment/envs/l5_env.py
```diff
@@ -0,0 +1,260 @@
...
+        self.cle = cle
+        self.rescale_action = rescale_action
+
+    def reset(self) -> Dict[str, np.ndarray]:
+        """ Resets the environment and outputs first frame of a new scene sample.
+
+        :return: the observation of first frame of sampled scene index
+        """
+        # Sample a episode randomly
+        self.scene_index = random.randint(0, self.max_scene_id)
```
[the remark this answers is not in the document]
theDebugger811: `init` helps to initialize the environment. `reset` is used to start a new episode (by sampling a new scene to be rolled out). The attributes defined in `reset` are specific to that particular rollout, hence not a part of `init`. Should I still define these attributes in `init`?

l5kit/l5kit/environment/reward.py
```diff
@@ -0,0 +1,88 @@
...
+
+    def get_reward(self, frame_index: int, simulated_outputs: List[SimulationOutputCLE]) -> float:
+        """Get the reward for the given step in close loop training.
+
+        :param frame_index: the frame index for which reward is to be calculated
+        :param simulated_outputs: the object contain the ego target and prediction attributes
+        :return: the reward is the combination of L2 error from groundtruth trajectory and (optionally) yaw error
+        """
+        scene_id = simulated_outputs[0].scene_id
+        self.metric_set.evaluate(simulated_outputs)
```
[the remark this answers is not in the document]
theDebugger811: Yes, I wanted to make use of existing code as much as possible. The distance error metrics support the compute for entire scene only. Do you think it is better to add an extra method in the `DisplacementErrorMetric` class to support per frame evaluation? It wont save a lot of compute I believe.

l5kit/l5kit/environment/reward.py
```diff
@@ -0,0 +1,88 @@
...
+
+        scene_metrics = self.metric_set.evaluator.scene_metric_results[scene_id]
+        dist_error = scene_metrics['displacement_error_l2'][frame_index + 1]
+        yaw_error = self.yaw_weight * scene_metrics['yaw_error_ca'][frame_index + 1]
+
+        # clip reward
+        reward = float(-dist_error.item())
+        if self.enable_clip:
```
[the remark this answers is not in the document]
theDebugger811: We just want to clip the distance error (in x, y) and not the yaw error (it is bounded).

l5kit/l5kit/environment/reward.py
```diff
@@ -0,0 +1,88 @@
...
+        """Return the reward at a particular time-step during the episode.
+
+        :param frame_index: the frame index for which reward is to be calculated
+        :param simulated_outputs: the object contain the ego target and prediction attributes
+        :return: reward at a particular frame index (time-step) during the episode.
+        """
+        raise NotImplementedError
+
+
+class CLE_Reward(Reward):
```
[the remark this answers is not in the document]
theDebugger811: It's not technically supervised right? There is not back-propagation of penalty (-reward). Removed the `_` from class name.

examples/RL/create_gym_example.py
```diff
@@ -0,0 +1,12 @@
+import os
```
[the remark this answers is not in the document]
theDebugger811: Yes, for now it's a file. It will be replaced with a notebook (TODO).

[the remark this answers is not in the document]
lucabergamini: add the TODO in the code then

l5kit/l5kit/environment/envs/l5_env.py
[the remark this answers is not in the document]
lucabergamini: yeah, but the name does not, there is no relation between `gym` and `bokeh`

l5kit/l5kit/environment/reward.py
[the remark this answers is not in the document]
lucabergamini: ok, add it as a comment

l5kit/l5kit/environment/reward.py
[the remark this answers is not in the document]
lucabergamini: I'm a little bit worried about this, for the following reason:
- we're passing a metricset to just use `displacement_error_l2` and `yaw_error_closest_angle`. However, we're not checking if those metrics are in the metricset;
- we're hiding how this reward is computed, as it can only be understood by reading the `get_reward` carefully.

A partial solution could be:
- assert what we need from the metricset in the init (I'm still not convinced it is worth passing the full metricset, maybe we could pass the two metrics)

regarding calling this over the whole scene:
- why not slicing it before passing it to the evaluate instead that afterwards?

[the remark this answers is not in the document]
theDebugger811: @perone what would you suggest?

l5kit/l5kit/environment/envs/l5_env.py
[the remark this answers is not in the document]
lucabergamini: better :)

l5kit/l5kit/cle/metric_set.py
lucabergamini: what is the purpose of this class?

l5kit/l5kit/simulation/unroll.py
lucabergamini: A bit. It seems we're introducing a lot of complexity though

l5kit/l5kit/cle/metric_set.py
theDebugger811: Its similar in spirit to the driveformer metric_set. [Ref](https://github.com/lyft/avjadoo/blob/master/experiments/planning/driveformer_cle_metricset.py)
Its a metric set for L5Kit. 

We can have different metric sets inheriting from this. Currently, we have one with distance and yaw error. In future, we can have one with collision metrics as well.

l5kit/l5kit/simulation/unroll.py
theDebugger811: Without this flag, we might have to end up writing a class very similar to ClosedLoopSimulator, but without the ego and agent models as the model (policy) in L5Gym is decoupled from the simulator (the environment).

I believe the verification flag is a small added complexity that allows to re-use the entire ClosedLoopSimulator for L5 gym environment. What do you think?

l5kit/l5kit/environment/envs/l5_env.py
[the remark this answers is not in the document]
theDebugger811: Thanks, I will then update the class name.

l5kit/l5kit/cle/metric_set.py
lucabergamini: got it.

It's a little bit weird because this class does nothing so we could also move everything in the super. But I guess we need this hierarchy for inheritance at some point

l5kit/l5kit/simulation/unroll.py
lucabergamini: should we have this more explicit? 
what about:
- defining an IntEnum with 2 modes (gym, l5kit)
- taking that as a param and perform the check based on that

what do you think?

theDebugger811: Thats certainly a better way. I will do this!

theDebugger811:
Thanks a lot @lucabergamini for the review and suggestions. 

@perone , @szagoruyko  I have incorporated Luca's review and previous comments by perone. Could you have another pass on the latest commit? Thanks!

l5kit/l5kit/environment/envs/l5_env.py
```diff
@@ -0,0 +1,323 @@
...
+        :return: the observation of first frame of sampled scene index
+        """
+        # Define in / outs for new episode scene
+        self.agents_ins_outs: DefaultDict[int, List[List[UnrollInputOutput]]] = defaultdict(list)
+        self.ego_ins_outs: DefaultDict[int, List[UnrollInputOutput]] = defaultdict(list)
+
+        if self.reset_scene_id is not None:
+            self.scene_index = self.reset_scene_id
+        else:  # Sample a scene
+            self.scene_index = random.randint(0, self.max_scene_id)
```
szagoruyko, line 181: you need to use `self.np_random.randint` here. `random.randint` does not have the seed set.

szagoruyko (commented):
started reviewing, going to continue tomorrow

l5kit/l5kit/environment/envs/l5_env.py
```diff
@@ -0,0 +1,323 @@
...
+                                     UnrollInputOutput)
+
+
+class SimulationConfigGym(SimulationConfig):
+    """Defines the default parameters used for the simulation of ego and agents around it in L5Kit Gym.
+
+    :param eps_length: the number of step to simulate per episode in the gym environment.
+    """
+
+    def __new__(cls, eps_length: int = 32, start_frame_idx: int = 0) -> 'SimulationConfigGym':
```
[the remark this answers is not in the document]
theDebugger811: I read online: [Ref](https://stackoverflow.com/questions/42385916/inheriting-from-a-namedtuple-base-class/42386174#42386174). It says we need to override __new__ in case of namedTuples.

l5kit/l5kit/environment/utils.py
```diff
@@ -0,0 +1,116 @@
...
+    for k, v in data.items():
+        if isinstance(v, int) or isinstance(v, float) or isinstance(v, np.int64) or isinstance(v, np.float32):
+            output_data[k] = np.array([v])
+        elif isinstance(v, np.ndarray):
+            output_data[k] = np.expand_dims(v, axis=0)
+        elif isinstance(v, torch.Tensor):
+            output_data[k] = np.expand_dims(v.cpu().numpy(), axis=0)
+        else:
+            print(k, v, type(v))
+            raise NotImplementedError
```
szagoruyko, line 83: best to avoid prints in the code for proper error handling
```suggestion
            raise NotImplementedError(f"{type(v)} is not supported (field {k})")
```

l5kit/l5kit/environment/envs/l5_env.py
```diff
@@ -0,0 +1,323 @@
...
+        self.reward.reset()
+
+        # Output first observation
+        self.frame_index = 1  # Frame_index 1 has access to the true ego speed
+        ego_input = self.sim_dataset.rasterise_frame_batch(self.frame_index)
+        self.ego_input_dict = convert_to_numpy(ego_input[0])
+
+        # Reset Kinematic model
+        if self.use_kinematic:
+            init_kin_state = np.array([0.0, 0.0, 0.0, 0.1 * ego_input[0]['curr_speed']])
```
szagoruyko, line 194: why multiply by 0.1?

szagoruyko (commented):
added few more comments

l5kit/l5kit/environment/envs/l5_env.py
theDebugger811: Because the step_time is 0.1. I have added a comment there.

examples/RL/create_gym_example.py
```diff
@@ -0,0 +1,19 @@
...
+
+import gym
+import l5kit.environment
+
+
+# TODO: Create a Google colab notebook for environment creation and training.
+
+# By setting the L5KIT_DATA_FOLDER variable, we can point the script
+# to the folder where the data lies.
+os.environ["L5KIT_DATA_FOLDER"] = os.environ["HOME"] + '/level5_data'
```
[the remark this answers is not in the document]
theDebugger811: Hi, it seems I keep missing the point what to do here. Could you please suggest an alternative? Thanks!

l5kit/l5kit/environment/__init__.py
```diff
@@ -0,0 +1,9 @@
+import gym
+
+
+# Register L5 Env
+gym.envs.register(
+    id='L5-CLE-v0',
+    entry_point="l5kit.environment.envs.l5_env:L5Env",
+    kwargs={'env_config_path': '../../../examples/RL/config.yaml'},
```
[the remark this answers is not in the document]
theDebugger811: Could you suggest a way to handle this error? This will also solve the previous comment regarding "having a default for `env_config_path` to be the `config.yaml `

l5kit/l5kit/environment/envs/l5_env.py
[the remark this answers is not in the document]
theDebugger811: Assigning `use_ego_gt: bool = False` before `__new__` leads to an error. I have defined the default arguments inside `__new__`

l5kit/l5kit/environment/utils.py
```diff
@@ -0,0 +1,116 @@
...
+
+def calculate_rescale_params(sim_dataset: SimulationDataset, use_kinematic: bool = False) -> ActionRescaleParams:
+    """Calculate the action un-normalization parameters from the simulation dataset.
+
+    :param sim_dataset: the input dataset to calculate the action rescale parameters
+    :param use_kinematic: flag to use the kinematic model
+    :return: the unnormalized action
+    """
+    if use_kinematic:
+        return ActionRescaleParams()
```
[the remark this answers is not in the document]
theDebugger811: `ActionRescaleParams()` is a namedTuple that defines the rescale parameters for both kinematic and non-kinematic params. Do you think having separate tuples `KinematicActionRescaleParams()` and `NonKinematicActionRescaleParams()` will help?

examples/RL/create_gym_example.py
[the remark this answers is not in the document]
perone: Just remove this line and change the comment to tell that the env variable L5KIT_DATA_FOLDER should be filled with the correct path for the download dataset.

l5kit/l5kit/environment/envs/l5_env.py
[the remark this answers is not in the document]
perone: Sorry I wasn't very clear, after doing this you should remove the `__new__` as you will have already defined the params, right ?

l5kit/l5kit/environment/utils.py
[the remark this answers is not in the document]
perone: What I meant was to keep the same named tuple, but just the way you create it should be changed. If the default params are linked to the kinematic params, then instead create a function to return the `ActionRescaleParams` with the kinematic params.

l5kit/l5kit/environment/envs/l5_env.py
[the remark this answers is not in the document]
theDebugger811: I want to inherit from` SimulationConfig` here as other parts of the code (eg. `ClosedLoopSimulator` require type `SimulationConfig`.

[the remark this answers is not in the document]
perone: But my suggestion was to still keep inheriting from it right ? Not sure if I understood.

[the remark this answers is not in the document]
theDebugger811: On making the requested changes, I got the following error:
`TypeError: <lambda>() missing 5 required positional arguments: 'use_ego_gt', 'use_agents_gt', 'disable_new_agents', 'distance_th_far', and 'distance_th_close'`

I believe due to inheritance, SimulationConfig is called and it does not have default values set for the above entries.  What do you suggest?

examples/RL/create_gym_example.py
```diff
@@ -0,0 +1,19 @@
...
+env_config_path = "./config.yaml"
+
+# call
+env = gym.make('L5-CLE-v0', env_config_path=env_config_path)
```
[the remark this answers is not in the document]
theDebugger811: The default `env_config_path` is `None` so that environment can be registered (as per our discussion today).

I personally prefer to explicitly point out the config file being used for this example.

l5kit/l5kit/environment/envs/l5_env.py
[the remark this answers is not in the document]
perone: But we can add defaults in the SimulationConfig right ?

examples/RL/create_gym_example.py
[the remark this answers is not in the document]
perone: That's fine, thanks !

theDebugger811:
@szagoruyko @perone Thanks for your feedback. I have incorporated the changes. @perone could you have another pass? Thanks!

examples/RL/create_gym_example.py
```diff
@@ -0,0 +1,22 @@
...
+import l5kit.environment
+
+
+# TODO: Create a Google colab notebook for environment creation and training.
+
+# Before starting, please download the Lyft L5 Prediction Dataset 2020 and
+# follow the instructions (https://github.com/lyft/l5kit#download-the-datasets)
+# to correctly organise it.
+
+# Set the L5KIT_DATA_FOLDER path and uncomment the line below
```
perone, line 13: This comment is not correct, it is either "set the environment variable" or "uncomment line below", not both. Setting the env variable and then uncommenting the line below will replace what was set.

l5kit/l5kit/cle/metric_set.py
```diff
@@ -0,0 +1,86 @@
...
+    """This class is responsible for computing a set of metric parametrization for the L5Kit.
+
+    :param metric_prefix: this is a prefix that will identify the metric set being used.
+    """
+
+    def __init__(self, metric_prefix: str = "L5") -> None:
+        """Constructor method
+        """
```
perone, line 38:
```suggestion
```

l5kit/l5kit/cle/metrics.py
```diff
@@ -237,3 +237,42 @@ def compute(self, simulation_output: SimulationOutput) -> torch.Tensor:
...
+
+class YawErrorMetric(SupportsMetricCompute):
+    """Yaw error computes the difference between the
+    simulated trajectory yaw and the observed trajectory yaw.
+
+    :param error_function: error function to compute distance
+    """
+    metric_name = "yaw_error"
+
+    def __init__(self, error_function: error_functions.ErrorFunction) -> None:
```
perone, line 250: Can set the `error_function` here to be the default as the closest angle.

l5kit/l5kit/environment/envs/l5_env.py
```diff
@@ -0,0 +1,333 @@
...
+    The speed at start_frame_idx is always 0 (not indicative of the true current speed).
+    We therefore simulate the episode from (start_frame_idx + 1, start_frame_idx + eps_length + 1)
+    """
+    use_ego_gt: bool = False
+    use_agents_gt: bool = True
+    disable_new_agents: bool = False
+    distance_th_far: float = 30.0
+    distance_th_close: float = 15.0
+    num_simulation_steps: int = 33
+    start_frame_idx: int = 0
```
perone, line 36: Can you add docs here so Sphinx can capture and generate automatically ? The format here should be something like:

```
(...)
#: doc for the variable goes here
start_frame_idx: int = 0
```

Another alternative is to add it in the docstr above, feel free to check which one looks better after Sphinx renders it.

l5kit/l5kit/environment/reward.py
```diff
@@ -0,0 +1,119 @@
...
+                 enable_clip: bool = True, rew_clip_thresh: float = 15.0,
+                 use_yaw: Optional[bool] = True, yaw_weight: Optional[float] = 20.0) -> None:
+        """Constructor method
+        """
+        self.reward_prefix = reward_prefix
+        # Metric Set
+        self.metric_set = metric_set if metric_set is not None else L2DisplacementYawMetricSet()
+
+        # Verify that error metrics necessary for reward calculation are present in the metric set
+        metric_names = [metric.metric_name for metric in self.metric_set.evaluation_plan.metrics]
```
perone, line 56: There is already a `metrics_dict` in the `EvaluationPlan`, so you don't have to collect metric names here, just get the dict and then do a `metric_name not in dict` to check.

l5kit/l5kit/environment/reward.py
```diff
@@ -0,0 +1,119 @@
...
+
+        self.enable_clip = enable_clip
+        self.rew_clip_thresh = rew_clip_thresh
+
+    def reset(self) -> None:
+        """Reset the closed loop evaluator when a new episode starts.
+        """
+        self.metric_set.reset()
+
+    def slice_simulated_output(self, index: int,
```
perone, line 73: This looks like a static method, can you add the `staticmethod` decorator and remove the self ? As it is not using  `self` for anything.

l5kit/l5kit/environment/envs/l5_env.py
```diff
@@ -0,0 +1,333 @@
...
+
+        :param frame_index: the index of the frame which provides the observation
+        :param episode_over: flag to determine if the episode is over
+        :return: the observation corresponding to the frame index
+        """
+        if episode_over:
+            frame_index = 0  # Dummy final obs (when episode_over)
+
+        ego_input = self.sim_dataset.rasterise_frame_batch(frame_index)
+        self.ego_input_dict = convert_to_numpy(ego_input[0])
```
perone, line 276: Same comment here, the `rasterise_frame_batch` return type is `List[Dict[str, np.ndarray]]`, so it seems to be already a numpy type correct ?

l5kit/l5kit/visualization/visualizer/zarr_utils.py
```diff
@@ -224,12 +225,102 @@ def simulation_out_to_visualizer_scene(sim_out: SimulationOutput, mapAPI: MapAPI
...
+        agents_frame = agents_frames[frame_idx]
+        agents_frame = filter_agents_by_labels(agents_frame, agents_th)
+        frame_vis = _get_frame_data(mapAPI, frame, agents_frame, tls_frame)
+        trajectories = []
+
+        if has_ego_info:
+            # print(frame_idx)
+            ego_in_out = ego_ins_outs[frame_idx]
+            # print("Target at: ", frame_idx, ego_in_out.inputs['target_positions'], ego_in_out.inputs['target_yaws'])
+            # print("Output at: ", frame_idx, ego_in_out.outputs['positions'], ego_in_out.outputs['yaws'])
```
perone, line 305: Same here

perone (commented):
Thanks @theDebugger811, added a few minor more comments.

examples/RL/create_gym_example.py
theDebugger811: Okay, just keeping "Uncomment the line below"

l5kit/l5kit/cle/metrics.py
theDebugger811: I am following the template of `DisplacementErrorMetric`. It doesn't have a default `error_function`. Can I keep as it is currently?

perone: For displacement error it makes more sense to have different metrics to compute, but for the yaw the one we're using it is a more natural option, users might confuse that and end up other metrics, so it should be ok to deviate here in this case.

theDebugger811: Okay, should I still keep the `YawErrorCAMetric` then?

l5kit/l5kit/environment/envs/l5_env.py
```diff
@@ -0,0 +1,333 @@
...
+            self.scene_index = self.np_random.randint(0, self.max_scene_id)
+        self.sim_dataset = SimulationDataset.from_dataset_indices(self.dataset, [self.scene_index], self.sim_cfg)
+
+        # Reset CLE evaluator
+        self.reward.reset()
+
+        # Output first observation
+        self.frame_index = 1  # Frame_index 1 has access to the true ego speed
+        ego_input = self.sim_dataset.rasterise_frame_batch(self.frame_index)
+        self.ego_input_dict = convert_to_numpy(ego_input[0])
```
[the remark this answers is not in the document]
perone: I think collate might be misleading here as it is not doing the same as the torch collate does. I wouldn't add another function just to expand the dimensions, can be done inline in the code as needed as it is a simple op.

theDebugger811:
@perone thanks for the quick review. I have incorporated the suggestions, removed the debug comments.

perone (approved):

PR #312, opened 2021-08-13
Fixing number generator seed

perone:
Patch by @qianguih.

### DIFF
--- a/l5kit/l5kit/random/random_generator.py
+++ b/l5kit/l5kit/random/random_generator.py
@@ -1,5 +1,5 @@
 from abc import ABC, abstractmethod
-from typing import Optional, Callable
+from typing import Callable, Optional
 
 import numpy as np
 
--- a/l5kit/l5kit/random/random_generator.py
+++ b/l5kit/l5kit/random/random_generator.py
@@ -1,5 +1,5 @@
 from abc import ABC, abstractmethod
-from typing import Callable
+from typing import Optional, Callable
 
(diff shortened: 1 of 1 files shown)


l5kit/l5kit/random/random_generator.py
```diff
@@ -58,12 +58,13 @@ class GaussianRandomGenerator(RandomGenerator):
     Note that this gaussian can be multidimensional.
     """
 
-    def __init__(self, mean: np.ndarray, std: np.ndarray):
+    def __init__(self, mean: np.ndarray, std: np.ndarray, random_seed: Optional[int] = 0):
```
lucabergamini, line 61: fine with this, although it should still be using the default numpy seed so it's not clear to me how this was creating any issues (multiproc maybe?)

lucabergamini (approved):

l5kit/l5kit/random/random_generator.py
perone: But there is a seeding for the workers in the dataloader too, so I'm not sure why was this causing issues too. cc @qianguih

qianguih: in the current implementation, the seed could be reset by any unexpected `np.random.seed` call. in general, it's safer to maintain it's own rng so that nothing else would affect it. check [this blog post](https://albertcthomas.github.io/good-practices-random-number-generators/) for more details.

PR #324, opened 2021-08-23
PPO policy and custom feature extractor added

theDebugger811:
The PR comprises:
1. A simple CNN backbone feature extractor with group norm (GN)
2. Custom Feature extractor for PPO using CNN+GN
3. Callbacks to save and evaluate model state
4. A script to run PPO on L5Kit.


theDebugger811:
@perone , @lucabergamini and @szagoruyko could you please review the PR which incorporates the PPO Policy into L5Kit? Thanks!

l5kit/l5kit/environment/ppo.py
```diff
@@ -0,0 +1,99 @@
...
+                    learning_rate=3e-4, gamma=args.gamma, tensorboard_log=args.tb_log, n_epochs=args.n_epochs,
+                    clip_range=clip_schedule, batch_size=args.batch_size, seed=args.seed, gae_lambda=args.gae_lambda)
+
+    # make eval env
+    eval_env_kwargs = {'env_config_path': args.config, 'use_kinematic': args.kinematic, 'return_info': True}
+    eval_env = make_vec_env("L5-CLE-v0", env_kwargs=eval_env_kwargs, n_envs=4,
+                            vec_env_cls=SubprocVecEnv, vec_env_kwargs=dict(start_method='fork'))
+
+    # callback list
+    callback = get_callback_list(eval_env, args.output, args.n_envs, save_freq=args.save_freq,
```
perone, line 95: There is just basically two callbacks in the `get_callback_list`, so let's move that function to be inside of this example instead of inside `environment.callbacks`, then the users will know what is inside of it as well.

l5kit/l5kit/environment/callbacks.py
```diff
@@ -0,0 +1,37 @@
+from typing import List, Optional
+
+import gym
+from stable_baselines3.common.callbacks import BaseCallback, CallbackList, CheckpointCallback, EvalCallback
+
+
+def get_callback_list(eval_env: gym.Env, output_prefix: str, n_envs: int,
```
perone, line 7: See my comment below about moving this function.

l5kit/l5kit/environment/callbacks.py
```diff
@@ -0,0 +1,37 @@
...
+    ckpt_prefix = ckpt_prefix if ckpt_prefix is not None else output_prefix
+    checkpoint_callback = CheckpointCallback(save_freq=(save_freq // n_envs), save_path=save_path,
+                                             name_prefix=ckpt_prefix)
+    callback_list.append(checkpoint_callback)
+
+    # Eval Model Periodically
+    eval_callback = EvalCallback(eval_env, eval_freq=(eval_freq // n_envs), n_eval_episodes=n_eval_episodes)
+    callback_list.append(eval_callback)
+
+    callback = CallbackList(callback_list)
```
perone, line 36: You can remove the `callback_list` variable and the appends and use `[checkpoint_callback, eval_callback]` here.

l5kit/l5kit/environment/callbacks.py
```diff
@@ -0,0 +1,37 @@
...
+    :param save_freq: the frequency to save the model
+    :param save_path: the folder to save the model
+    :param eval_freq: the frequency to evaluate the model
+    :param n_eval_episodes: the number of episodes to evaluate
+    :param ckpt_prefix: the prefix to save the model during training
+    """
+    callback_list: List[BaseCallback] = []
+
+    # Save Model Periodically
+    ckpt_prefix = ckpt_prefix if ckpt_prefix is not None else output_prefix
```
perone, line 27:
```suggestion
    ckpt_prefix = ckpt_prefix or output_prefix
```

l5kit/l5kit/environment/feature_extractor.py
```diff
@@ -0,0 +1,54 @@
...
+            # A simplified feature extractor with GroupNorm.
+            model = models.SimpleCNN_GN(num_input_channels, features_dim)
+        else:
+            raise NotImplementedError
+
+        extractors = {}
+        total_concat_size = 0
+        # We need to know size of the output of this extractor,
+        # so go over all the spaces and compute output feature sizes
+        for key, subspace in observation_space.spaces.items():
```
perone, line 35: Is `observation_space.spaces` a dict ? If yes, then it doesn't make much sense to iterate on it as you can just index to find the keys you need.

l5kit/l5kit/environment/ppo.py
```diff
@@ -0,0 +1,99 @@
...
+                        help='Reward clipping threshold')
+    parser.add_argument('--kinematic', action='store_true',
+                        help='Flag to use kinematic model in the environment')
+    parser.add_argument('--seed', default=42, type=int)
+    args = parser.parse_args()
+
+    # make train env
+    env_kwargs = {'env_config_path': args.config, 'use_kinematic': args.kinematic}
+    env = make_vec_env("L5-CLE-v0", env_kwargs=env_kwargs, n_envs=args.n_envs,
+                       vec_env_cls=SubprocVecEnv, vec_env_kwargs=dict(start_method='fork'))
```
perone, line 71: Use `{"start_method": "fork"}` instead of `dict()`

l5kit/l5kit/environment/ppo.py
```diff
@@ -0,0 +1,99 @@
...
+
+    # Custom Feature Extractor backbone
+    policy_kwargs = dict(
```
perone, line 74: Same here, use `{`

l5kit/l5kit/environment/ppo.py
```diff
@@ -0,0 +1,99 @@
...
+        features_extractor_class=CustomFeatureExtractor,
+        features_extractor_kwargs=dict(features_dim=128, model_arch=args.model_arch),
```
perone, line 76: Same here

l5kit/l5kit/environment/ppo.py
```diff
@@ -0,0 +1,99 @@
+import argparse
```
perone, line 1: This file is an example on how to use PPO, and it is also a script. Users install `l5kit` with `pip install l5kit`, so how do they get access to this script ? A better way to structure it would be to move this to the examples folder.

perone (changes requested):
Thanks @theDebugger811, added a few comments.

theDebugger811:
Thanks @perone for the quick review. I have incorporated the suggestions. Could you have another pass?

examples/RL/ppo.py
```diff
@@ -0,0 +1,107 @@
...
+from stable_baselines3.common.env_util import make_vec_env
+from stable_baselines3.common.utils import get_linear_fn
+from stable_baselines3.common.vec_env import SubprocVecEnv
+
+from l5kit.environment.feature_extractor import CustomFeatureExtractor
+
+
+os.environ["L5KIT_DATA_FOLDER"] = "/tmp/l5kit_data"
+# Dataset is assumed to be on the folder specified
+# in the L5KIT_DATA_FOLDER environment variable
```
perone, line 15: Move the comment block to above the code line.

examples/RL/ppo.py
```diff
@@ -0,0 +1,107 @@
...
+from stable_baselines3 import PPO
+from stable_baselines3.common.callbacks import CheckpointCallback, EvalCallback
+from stable_baselines3.common.env_util import make_vec_env
+from stable_baselines3.common.utils import get_linear_fn
+from stable_baselines3.common.vec_env import SubprocVecEnv
+
+from l5kit.environment.feature_extractor import CustomFeatureExtractor
+
+
+os.environ["L5KIT_DATA_FOLDER"] = "/tmp/l5kit_data"
```
perone, line 13: I think it would be better to check if this environment variable is present and then throw a warning, instead of assuming that the user will always put data into `/tmp/l5kit_data`. Also, if we want to call this example from another place and set this environment variable before it, it will be overwritten after it starts executing. I would leave for the user/caller of this script the responsibility of setting the `L5KIT_DATA_FOLDER`. What happens today if this is not set ? Is there any warning already ? If yes, then we can just remove it, if there is an error that is not clear, then we should check and throw a clear error message.

examples/RL/ppo.py
```diff
@@ -0,0 +1,107 @@
...
+
+    # make train env
+    env_kwargs = {'env_config_path': args.config, 'use_kinematic': args.kinematic}
+    env = make_vec_env("L5-CLE-v0", env_kwargs=env_kwargs, n_envs=args.n_envs,
+                       vec_env_cls=SubprocVecEnv, vec_env_kwargs={"start_method": "fork"})
+
+    # Custom Feature Extractor backbone
+    policy_kwargs = {
+        "features_extractor_class": CustomFeatureExtractor,
+        "features_extractor_kwargs": {"features_dim": 128, "model_arch": args.model_arch},
```
perone, line 76: Add the `features_dim` as a argparse parameter.

examples/RL/ppo.py
```diff
@@ -0,0 +1,107 @@
...
+        "normalize_images": False
+    }
+
+    # define model
+    clip_schedule = get_linear_fn(args.clip_start_val, args.clip_end_val, args.clip_progress_ratio)
+    if args.load is not None:
+        model = PPO.load(args.load, env, clip_range=clip_schedule)
+    else:
+        model = PPO("CnnPolicy", env, policy_kwargs=policy_kwargs, verbose=1, n_steps=args.num_rollout_steps,
+                    learning_rate=3e-4, gamma=args.gamma, tensorboard_log=args.tb_log, n_epochs=args.n_epochs,
```
perone, line 86: Same for `learning_rate` here, it is an hyper-param.

examples/RL/ppo.py
```diff
@@ -0,0 +1,107 @@
...
+                    clip_range=clip_schedule, batch_size=args.batch_size, seed=args.seed, gae_lambda=args.gae_lambda)
+
+    # make eval env
+    eval_env_kwargs = {'env_config_path': args.config, 'use_kinematic': args.kinematic, 'return_info': True}
+    eval_env = make_vec_env("L5-CLE-v0", env_kwargs=eval_env_kwargs, n_envs=4,
```
perone, line 91: Same for `n_envs=4`, add it to the argparse params.

perone (changes requested):
Just a few minor comments, thanks @theDebugger811 !

examples/RL/ppo.py
theDebugger811: I have added the following check:
```
if "L5KIT_DATA_FOLDER" not in os.environ:
    raise KeyError("L5KIT_DATA_FOLDER environment variable not set")

```

theDebugger811:
Thanks @perone , comments incorporated.

perone (approved):
Thanks @theDebugger811 !

PR #325, opened 2021-08-24
Evaluation protocol for RL environment

theDebugger811:
Adding evaluation protocol for RL environment to measure generalization capabilities.

### DIFF
--- a/examples/RL/ppo.py
+++ b/examples/RL/ppo.py
@@ -8,10 +8,12 @@ from stable_baselines3.common.utils import get_linear_fn
 from stable_baselines3.common.vec_env import SubprocVecEnv
 
 from l5kit.environment.feature_extractor import CustomFeatureExtractor
+from l5kit.environment.callbacks import L5KitEvalCallback
 
 # Dataset is assumed to be on the folder specified
 # in the L5KIT_DATA_FOLDER environment variable
 # Please set the L5KIT_DATA_FOLDER environment variable
+os.environ["L5KIT_DATA_FOLDER"] = os.environ["HOME"] + '/level5_data/'
 if "L5KIT_DATA_FOLDER" not in os.environ:
     raise KeyError("L5KIT_DATA_FOLDER environment variable not set")
 
@@ -107,8 +109,10 @@ if __name__ == "__main__":
                                              name_prefix=args.output)
 
     # Eval Model Periodically
-    eval_callback = EvalCallback(eval_env, eval_freq=(args.eval_freq // args.n_envs),
-                                 n_eval_episodes=args.n_eval_episodes)
+    # eval_callback = EvalCallback(eval_env, eval_freq=(args.eval_freq // args.n_envs),
+    #                              n_eval_episodes=args.n_eval_episodes)
+    eval_callback = L5KitEvalCallback(eval_env, eval_freq=(args.eval_freq // args.n_envs),
+                                      n_eval_episodes=args.n_eval_episodes, n_eval_envs=args.n_eval_envs)
 
     # train
     model.learn(args.n_steps, callback=[checkpoint_callback, eval_callback])
--- a/l5kit/l5kit/environment/callbacks.py
+++ b/l5kit/l5kit/environment/callbacks.py
@@ -1,0 +1,75 @@
+from typing import List
+
+import gym
+from stable_baselines3.common.callbacks import EvalCallback
+
+from l5kit.cle.validators import ValidationCountingAggregator
+from l5kit.environment.envs.l5_env import EpisodeOutputGym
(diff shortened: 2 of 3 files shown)


theDebugger811:
@perone, could you let me know what would be ideal to plot on tensorboard for evaluation post the metric and validator compute.
Ref: [See here](https://github.com/lyft/l5kit/blob/3581f467cbc97b4735c3e8af62597f542ada9e58/l5kit/l5kit/environment/callbacks.py#L43)

Thanks!

examples/RL/ppo.py
```diff
@@ -8,10 +8,12 @@
 from stable_baselines3.common.vec_env import SubprocVecEnv
 
 from l5kit.environment.feature_extractor import CustomFeatureExtractor
+from l5kit.environment.callbacks import L5KitEvalCallback
 
 # Dataset is assumed to be on the folder specified
 # in the L5KIT_DATA_FOLDER environment variable
 # Please set the L5KIT_DATA_FOLDER environment variable
+os.environ["L5KIT_DATA_FOLDER"] = os.environ["HOME"] + '/level5_data/'
```
perone, line 16: Remember to remove this as it is assuming that users are putting their data at `~/level5_data`

examples/RL/ppo.py
```diff
@@ -107,8 +109,10 @@
                                              name_prefix=args.output)
 
     # Eval Model Periodically
-    eval_callback = EvalCallback(eval_env, eval_freq=(args.eval_freq // args.n_envs),
-                                 n_eval_episodes=args.n_eval_episodes)
+    # eval_callback = EvalCallback(eval_env, eval_freq=(args.eval_freq // args.n_envs),
+    #                              n_eval_episodes=args.n_eval_episodes)
```
perone, line 113: Remove the old comment

l5kit/l5kit/environment/callbacks.py
```diff
@@ -0,0 +1,75 @@
...
+        if self.eval_freq > 0 and self.n_calls % self.eval_freq == 0:
+            # Get episode outputs
+            episode_outputs = self.get_eval_info()
+
+            # Evaluate
+            self.metric_set.evaluator.evaluate(episode_outputs)
+            validation_results = self.metric_set.evaluator.validation_results()
+            _ = ValidationCountingAggregator().aggregate(validation_results)
+
+            # TODO Log the aggregated results
```
perone, line 43: You can add all metrics we have available for the CLE in l5kit.

theDebugger811: I wanted to ask what is usually the case? e.g.
1. All the `metrics`? Averaged across time or measured at 'last' time step?
2. All the `range validators`?
3. Both?

perone: We already have the `ValidationCountingAggregator` there, so you can get from this aggregation (it will aggregate per scene), at the end of the day we want to know for the entire number of scenes, what were the results aggregated (how many collisions in total, for the entire amount of scenes, for a particular policy)

theDebugger811:
@perone, logged the results of `ValidationCountingAggregator` into tensorboard.

l5kit/l5kit/environment/callbacks.py
theDebugger811: Done!

l5kit/l5kit/environment/gym_metric_set.py
```diff
@@ -23,3 +23,50 @@ def build_metrics(self) -> List[metrics.SupportsMetricCompute]:
             metrics.DisplacementErrorL2Metric(),
             metrics.YawErrorCAMetric()
         ]
+
+
+class DisplacementCollisionMetricSet(metric_set.L5MetricSet):
```
perone, line 28: The name of this class is a little confusing, it says DisplacementCollisionMetricSet but we also have distance to ref traj here. I would just change it to be something like CLEMetricSet.

l5kit/l5kit/environment/callbacks.py
```diff
@@ -0,0 +1,70 @@
...
+    """
+
+    def __init__(self, eval_env: gym.Env, eval_freq: int = 10000, n_eval_episodes: int = 10,
+                 n_eval_envs: int = 4, verbose: int = 0) -> None:
+        super(L5KitEvalCallback, self).__init__(eval_env)
+        self.eval_freq = eval_freq
+        self.n_eval_episodes = n_eval_episodes
+        self.n_eval_envs = n_eval_envs
+        self.verbose = verbose
+        self.metric_set = DisplacementCollisionMetricSet()
```
perone, line 25: If the user wants to change this, it will have to extend this class and change the constructor. Another alternative is to receive this a a parameter.

PR #328, opened 2021-09-08
Scene categorization and scene type evaluation

theDebugger811:
PR comprises:
1. Script to categorize zarr_dataset based on ego turning.
2. Provide metadata for validation set
3. Evaluation based on different scene types.
4. Minor docs update: Remove email link from docs.


scripts/categorize_scenes.py
```diff
@@ -0,0 +1,92 @@
...
+
+        # Update dict
+        turn_dict[scene_id] = turn_type
+    return turn_dict
+
+
+# Dataset is assumed to be on the folder specified
+# in the L5KIT_DATA_FOLDER environment variable
+# Please set the L5KIT_DATA_FOLDER environment variable
+os.environ["L5KIT_DATA_FOLDER"] = "/home/ubuntu/level5_data"
```
perone, line 65: Do not set here, users should do it w/ the environment variable.

l5kit/l5kit/environment/callbacks.py
```diff
@@ -16,21 +18,50 @@ class L5KitEvalCallback(EvalCallback):
...
+        self.enable_scene_type_aggregation = enable_scene_type_aggregation
+        if self.enable_scene_type_aggregation:
+            assert scene_id_to_type_path is not None
+            self.scene_id_to_type_path = scene_id_to_type_path
+            self.scene_ids_to_scene_types = self._get_scene_types()
+
     def _init_callback(self) -> None:
         pass
 
+    def _get_scene_types(self, path: Optional[str] = None) -> List[List[str]]:
```
perone, line 47: Something is weird here, the `path` is not being used.

l5kit/l5kit/environment/callbacks.py
```diff
@@ -16,21 +18,50 @@ class L5KitEvalCallback(EvalCallback):
...
+        scene_type_dict: Dict[int, str]
+        with open(self.scene_id_to_type_path, 'r') as f:
+            csv_reader = csv.reader(f)
+            scene_type_dict = {int(rows[0]): rows[1] for rows in csv_reader}
+
+        # Convert dict to List[List[str]]
+        scene_id_to_type_list: List[List[str]] = []
+        for k, v in scene_type_dict.items():
+            scene_id_to_type_list.append([v])
+        return scene_id_to_type_list
```
perone, line 63: Would a dict make more sense here ?

scripts/categorize_scenes.py
```diff
@@ -0,0 +1,92 @@
...
+    # load dataset
+    dm = LocalDataManager()
+    dataset_path = dm.require(args.data_path)
+    zarr_dataset = ChunkedDataset(dataset_path)
+    zarr_dataset.open()
+
+    # categorize
+    turn_dict = identify_turn(zarr_dataset)
+    res = Counter(turn_dict.values())
+    print(res)
```
perone, line 86: What is being printed here ?

scripts/categorize_scenes.py
```diff
@@ -0,0 +1,92 @@
+import argparse
```
perone, line 1: Add a small readme in the folder to mention how to use the script and what they do.

scripts/validate_turns_metadata.csv
```diff
@@ -0,0 +1,16220 @@
+0,right
```
perone, line 1: Is this for the entire dataset or just for training/val ?

perone (commented):
Thanks @theDebugger811, added a few comments.

scripts/categorize_scenes.py
theDebugger811: I always forget to remove it from one place. :)

l5kit/l5kit/environment/callbacks.py
theDebugger811: The original `compute_cle_scene_type_aggregations` function assume it is in form `List[List[str]]`. Keeping the consistency.

scripts/validate_turns_metadata.csv
theDebugger811: Just for `val`. I will add two more for `train` and `train_full` as well.

theDebugger811:
Thanks @perone for the review. Let me know if the comments are addressed.

scripts/categorize_scenes.py
```diff
@@ -0,0 +1,91 @@
...
+    # load dataset
+    dm = LocalDataManager()
+    dataset_path = dm.require(args.data_path)
+    zarr_dataset = ChunkedDataset(dataset_path)
+    zarr_dataset.open()
+
+    # categorize
+    turn_dict = identify_turn(zarr_dataset)
+    categories_counter = Counter(turn_dict.values())
+    print(categories_counter)
```
perone, line 85: It seems to me that the user won't know what is this so it would be nice to add a print before this one telling what comes below.

scripts/categorize_scenes.py
```diff
@@ -0,0 +1,91 @@
...
+
```
perone, line 86: Make sure to move the `.csv` files to another folder as they are assets and not scripts. You can create a new folder called `dataset_metadata` and also add a readme inside of this folder explaining what are these files, how they were generated and to which dataset they were generated from.

perone (commented):
Thanks, just added a minor and another comment to reorganize the csv files outside of the `scripts` folder.

theDebugger811:
Hi @perone , reorganized the csv files. Let me know if it looks good now.

perone (approved):
Thanks !

PR #329, opened 2021-09-09
Scene aggregation default arg fixed

theDebugger811:


theDebugger811:
@perone could you approve this?

perone (approved):

PR #332, opened 2021-09-17
Introducing Vectorizer + vectorized Planning Model (Urban Driver)

oliver-scheel:
Add implementation from [Urban Driver](https://arxiv.org/abs/2109.13333) including:

- new `EgoDatasetVectorized` with data pipeline to get vectorized input representation;
- Urban Driver model definition (both final model and ablations);
- Training and evaluation notebooks for Urban Driver;
- Pre-trained models (both final model and ablations)


lucabergamini (approved):
LGTM

PR #338, opened 2021-09-24
Add urban planner notebooks

lucabergamini:
Add train and evaluation notebooks for urban planner. Features:

- train model and ablations;
- evaluate trained model in closed-loop eval with metrics and visualisation


oliver-scheel (approved):
Nice work, LGTM!

PR #342, opened 2021-09-29
Training Bug Fixes

oliver-scheel:
Fixes two bugs:
- multiplication of target weights with discount factor was wrong
- future_num_frames was always set to 3 as it was never written to the batch

### DIFF
--- a/l5kit/l5kit/planning/vectorized/closed_loop_model.py
+++ b/l5kit/l5kit/planning/vectorized/closed_loop_model.py
@@ -67,9 +67,7 @@ class VectorizedUnrollModel(VectorizedModel):
 
     def forward(self, data_batch: Dict[str, torch.Tensor]) -> Dict[str, torch.Tensor]:
         # ==== get additional info from the batch, or fall back to sensible defaults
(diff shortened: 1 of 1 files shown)


oliver-scheel:
meow @lucabergamini

l5kit/l5kit/planning/vectorized/closed_loop_model.py
```diff
@@ -270,7 +268,7 @@ def forward(self, data_batch: Dict[str, torch.Tensor]) -> Dict[str, torch.Tensor
 
             # discount timesteps t via discount_factor**t
             target_weights *= torch.tensor([self.discount_factor**(t - self.warmup_num_frames) for t in range(
-                target_weights.shape[0])])[..., None, None].to(target_weights.device)
+                target_weights.shape[1])])[None, ..., None].to(target_weights.device)
```
lucabergamini, line 271: this is complex, can we brake it in two lines and add shape comment?

lucabergamini (approved):
nits, LGTM

PR #345, opened 2021-10-06
Add DrivenMiles, Passiveness and total collision metrics

theDebugger811:


l5kit/l5kit/environment/callbacks.py
```diff
@@ -83,6 +83,17 @@ def _on_step(self) -> bool:
                 self.logger.record(f'{self.prefix}/ade', ade_error)
                 self.logger.record(f'{self.prefix}/fde', fde_error)
 
+            # Calculate replayed/simulated driven miles
+            if 'replay_driven_miles' in self.metric_set.evaluation_plan.metrics_dict():
```
perone, line 87: I think we should check if we have these keys during the construction of this callback and trigger an error if the metric set doesn't have the required metrics/cms/etc.

l5kit/l5kit/environment/callbacks.py
```diff
@@ -83,6 +83,17 @@ def _on_step(self) -> bool:
...
+                replay_driven_miles, simulated_driven_miles = self.compute_driven_miles()
+                self.logger.record(f'{self.prefix}/replay_driven_miles', replay_driven_miles)
+                self.logger.record(f'{self.prefix}/simulated_driven_miles', simulated_driven_miles)
+
+            # Calculate total collisions
+            if 'collision_front' in self.metric_set.evaluation_plan.metrics_dict():
```
perone, line 93: same here

l5kit/l5kit/environment/callbacks.py
```diff
@@ -145,3 +156,34 @@ def compute_ade_fde(self) -> Tuple[float, float]:
         average_fde = sum(scene_fde_list) / len(scene_fde_list)
 
         return (average_ade, average_fde)
+
+    def compute_driven_miles(self) -> Tuple[float, float]:
```
perone, line 160: We do have a composite metric for computing driven miles, why not using that one ?

perone (commented):
Thanks @theDebugger811, just a few comments.

l5kit/l5kit/environment/callbacks.py
theDebugger811: Done!

perone (approved):

PR #346, opened 2021-10-06
Test & fix L5Kit notebooks

eeshakumar:
- Fixed Planning Model import exception
- Output of default_collate is not sent to device, so steps were added to send batch data to device

### DIFF
--- a/examples/planning/open_loop_test.ipynb
+++ b/examples/planning/open_loop_test.ipynb
@@ -2,7 +2,6 @@
  "cells": [
   {
    "cell_type": "markdown",
-   "metadata": {},
    "source": [
     "# Open-Loop Evaluation\n",
     "\n",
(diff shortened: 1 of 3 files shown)


examples/planning/open_loop_test.ipynb
```diff
@@ -229,37 +227,38 @@
...
    "source": [
     "for frame_number in range(0, len(eval_dataset), len(eval_dataset) // 20):\n",
     "    \n",
     "    data = eval_dataloader.dataset[frame_number]\n",
-    "\n",
     "    data_batch = default_collate([data])\n",
-    "    \n",
+    "    for key, val in data_batch.items():\n",
+    "        if isinstance(val, torch.Tensor):\n",
+    "            data_batch[key] = val.to(device)\n",
```
lucabergamini, line 261: isn't enough to do 
```python
data_batch = {k: v.to(device) for k, v in data_batch.items()}
```

here? everything should be tensor here as far as I remember

lucabergamini (commented):
1 nit but generally LGTM

examples/planning/open_loop_test.ipynb
eeshakumar: Yes, that also works. I've changed it.

lucabergamini (approved):

PR #364, opened 2021-11-11
Add Simnet mode to DriverGym

theDebugger811:


theDebugger811:
Hi @perone, ptal! :D

l5kit/l5kit/environment/envs/l5_env.py
```diff
@@ -152,8 +155,14 @@ def __init__(self, env_config_path: Optional[str] = None, dmg: Optional[LocalDat
 
         # Simulator Config within Gym
         self.sim_cfg = sim_cfg if sim_cfg is not None else SimulationConfigGym()
+        simulation_model = None
+        if not self.sim_cfg.use_agents_gt:
+            self.device = torch.device("cpu")
+            simulation_model = torch.jit.load(simnet_model_path).to(torch.device("cpu"))
```
perone, line 161: It seems to have repeated `torch.device()` here. Also, is this the device where SimNet is going to do inference ? Why is it fixed to cpu ?

perone (commented):
Thanks @theDebugger811, added a comment regarding a device only.

l5kit/l5kit/environment/envs/l5_env.py
theDebugger811: Fixed the `device` attribute. The device is fixed to CPU as SimNet is part of the environment rollouts that take place on the CPU (GPU handles the RL policy training). Else there would be too much communication and hence copy between CPU and GPU (if device = "gpu").

perone: I see, but it will also be much slower w/ SimNet in CPU right ? Even though the GPU is for the training, during the rollouts, when SimNet is doing inference, the GPU is idle because the rollout need the results from SimNet to do the step right ? But let's merge this, just adding a comment for future reference.

perone (approved):

PR #417, opened 2023-06-02
Fix naming and dataset URLs

stefanopini:
Fix naming and dataset URLs after Woven Planet's name change to Woven by Toyota (April 1st, 2023) and new website at [woven.toyota](https://woven.toyota).


perone (approved):
Thanks !

PR #422, opened 2023-07-20
Restore filenames of downloaded data

stefanopini:
Restore filenames of downloaded data to keep compatibility with the rest of the codebase.


perone (approved):
💯

